Issue #66, opened 2026-01-29
Manual Crop Innaccurate when Fine Rotation is Applied


rodg:
## System/Version Info
OS: Rocky Linux, KDE Plasma Wayland
NegPy Version: Current `main` branch as of writing

## Issue

When applying a fine rotation, if you go to manually crop the image afterwards the crop will be misplaced. 

Image with no settings changed:
<img width="1935" height="1099" alt="Image" src="https://github.com/user-attachments/assets/c6017817-262f-4b02-8535-2d7b152067ba" />

Rotation Applied:
<img width="1935" height="1099" alt="Image" src="https://github.com/user-attachments/assets/358d3928-4dff-4100-8d7f-e0b67ccea17a" />

Crop Applied:
<img width="1935" height="1099" alt="Image" src="https://github.com/user-attachments/assets/d9f62e2d-1a8c-417a-8c8f-f7c108678979" />

When manually cropping, I am certain I'm excluding the white boarders. If I do the process without rotating, the crop works as expected. Looking through the code, I noticed that with manual crop + fine rotation, `cv2.getRotationMatrix2D` is being called on the image twice. That might have something to do with it? I might also be misunderstanding how the fine rotation is supposed to work.

rodg:
Messed with it a bit, was wrong about that `cv2.getRotationMatrix2D` I think. This actually only occurs when you enable GPU acceleration, so it must be something in the GPU logic. I also noticed that the direction of rotation changes depending on which rendering mode you're in.

PR #67, opened 2026-01-29
Fix GPU Fine Image Rotation

rodg:
Relating to #66. For fixing the rotation direction, I just tried reversing the direction and it fixed it. I then noticed that it looked a bit different in GPU vs CPU mode, and realized it was b/c for GPU accelerated fine rotation we weren't taking into account the image aspect ratio. I'm not a GPU expert so there may be a better way to make these changes. 

## Changes
- Fine image rotation now goes the same direction for both GPU and CPU 
- GPU fine image rotation now takes image aspect ratio into consideration

## Testing
I tested this on Rocky/KDE with an NVIDIA GPU. I put a few images through it and it all seemed to work more consistently.


marcinz606:
nice, i'll test it over the weekend!

Issue #146, opened 2026-05-01
Canvas color doesn't work without GPU acceleration


kosma:
Canvas color doesn't work without GPU acceleration - always stays black. The color is correct when GPU acceleration is enabled.

Version: 0.14.2
Platform: Windows 11

PR #161, opened 2026-05-04
feat: Improve autocrop behavior and canvas background handling

alessandrv:
- Reused personal project to get an improved autocrop detection using stronger contour-based frame detection with a threshold-based fallback.
- Made the Auto crop action set the crop ratio to Free.
- Made auto crop not always active, I feel like it was not really intiuitive having it always on, so autocrop now actually auto crops rather than just resetting the crop.
- Made manual crop disable the active Auto crop state once a manual crop is committed.
- Kept the Auto button visually highlighted while autocrop is active.

- Fixed canvas background rendering for CPU / non-GPU rendering paths by drawing the selected color correctly in the overlay path on windows and on non gpu accelerated instanced (not sure if it affected also linux while using CPU). #146 
- Updated tests to cover explicit autocrop enablement, manual-crop state transitions, and geometry logic behavior

Notes:
My auto crop implementation could be improved upon still but from my negative scans it feels a lot better than the previous implementation even if it fails sometimes but it happened also before on those frames. Might be relevant to allow a bigger negative crop offset as the current implementation cuts on the image borders.

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -49,6 +49,7 @@ class AppController(QObject):
     loading_started = pyqtSignal()
     export_progress = pyqtSignal(int, int, str)
     export_finished = pyqtSignal(float)
+    preview_loaded = pyqtSignal()
     render_requested = pyqtSignal(RenderTask)
     preview_load_requested = pyqtSignal(PreviewLoadTask)
     normalization_requested = pyqtSignal(NormalizationTask)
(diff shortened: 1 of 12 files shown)


marcinz606:
looks good, I've pushed one small commit that allows to resets crop simply by unselecting autocrop button.

marcinz606:
I plan to add few more things around the weekend, then your changes will be in 0.18.0 :) thanks for contributing!

PR #209, opened 2026-05-18
perf: Optimize preview demosaic

reederphill:
### Summary

Preview loading was using the same high-quality demosaic algorithms as the export pipeline, even though preview images are downsampled to 1600px max — a ~13× linear reduction where quality differences between algorithms are invisible. Additionally, a Numba JIT cold-start cost (~0.19s) was silently assigned to whichever format loaded first (always CR2 due to fixture ordering).

### Root causes

1. **Wrong algorithm for the context** — `get_best_demosaic_algorithm` returned AHD (Bayer) and VNG (X-Trans) for both preview and export. At preview scale, AHD and VNG are quality overkill with real runtime cost: AHD is ~60% slower than PPG for Bayer, VNG is ~4.6× slower than LINEAR for X-Trans.

2. **Numba JIT cold-start on first preview** — `uint16_to_float32` (used in `load_linear_preview`) is JIT-compiled by Numba. The first call in any process pays a ~0.19s overhead. In the metrics test this was fully attributed to CR2 (fixture position 0), making it appear dramatically slower than it actually is.

### Changes

**`negpy/infrastructure/loaders/helpers.py`**
- Added `for_preview: bool = False` parameter to `get_best_demosaic_algorithm`
- Preview mode returns PPG (Bayer) and LINEAR (X-Trans); default behaviour (AHD/VNG) is unchanged for export

**`negpy/services/rendering/preview_manager.py`**
- Passes `for_preview=True` when calling `get_best_demosaic_algorithm`
- Adds a module-level JIT warmup (2×2 dummy array) so the first real preview load doesn't pay the compile cost

### Performance

Measured on Apple M-series (arm64), cold load, `preview_render_size=1600`:

| Format | Before (MPx/s) | After (MPx/s) | Speedup |
|--------|---------------|--------------|---------|
| RAF (X-Trans) | 8.7 | ~52 | **5.9×** |
| CR2 (Canon) | 9.8 | ~25 | **2.5×** |
| DNG | 15.5 | ~32 | 2.1× |
| NEF | 16.6 | ~32 | 1.9× |
| ARW | 20.8 | ~53 | 2.5× |

### Quality impact

None for the preview path. At 1600px output from a 22 MPx sensor:
- PPG vs AHD on Bayer: indistinguishable after `cv2.INTER_AREA` downsampling
- LINEAR vs VNG on X-Trans: artifacts smoothed out by the ~3× linear downscale

Export path is unaffected — it still uses AHD and VNG.

### Testing

- 354 unit/integration tests pass (`pytest tests/ --ignore=tests/metrics`)
- Metrics test passes for all five fixture formats

### DIFF
--- a/negpy/infrastructure/loaders/helpers.py
+++ b/negpy/infrastructure/loaders/helpers.py
@@ -117,10 +117,13 @@ class NonStandardFileWrapper:
         return (data * 255.0).astype(np.uint8)
 
 
-def get_best_demosaic_algorithm(raw: Any) -> Any:
+def get_best_demosaic_algorithm(raw: Any, for_preview: bool = False) -> Any:
     """
     Selects optimal demosaicing algorithm based on sensor type and CFA pattern.
     Exclusively uses algorithms packaged in the standard permissive (LGPL) rawpy build.
+
+    When for_preview=True, selects faster algorithms appropriate for downsampled
+    preview rendering (PPG for Bayer, LINEAR for X-Trans).
     """
     selected_algo = rawpy.DemosaicAlgorithm.LINEAR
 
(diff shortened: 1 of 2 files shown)


reederphill:
This is my "barebones" improvement for RAF loading. Compared to https://github.com/marcinz606/NegPy/pull/210 it barely does anything, but if 210 is a bit too aggressive of a change this would be a good fallback.

PR #267, opened 2026-06-20
feat: polygon-mask dodge/burn local adjustments

reederphill:
First go at adding dodge/burn mechanics. Tried a brush tool but really hated the UX and I find the polygon masking more accurate and efficient.

Rendering runs as a new cached pipeline stage after lab/before toning, falling back to CPU when masks are present since the GPU engine doesn't yet implement this stage.


marcinz606:
nice, I also tried implementig it using brush tool and wasn't happy with results so i scrapped the idea :) will test it out

marcinz606:
i resolved conflicts & added some unit tests on separate branch https://github.com/marcinz606/NegPy/pull/274
will close this one

PR #281, opened 2026-06-21
Add log-style flat master export

### LINKED ISSUE #278: Feature request: Flat neutral positive / log-style export for further editing
As an end user, I would like an option to convert a negative into a very flat, neutral positive image intended as a digital intermediate for further editing in software such as Lightroom, Darktable or Photoshop.

The goal would not be to create a finished-looking image. Instead, the output should retain as much usable tonal and colour information as possible, with minimal creative interpretation.

Something conceptually similar to a Cineon, LogC or other log-style intermediate could be useful, although it would not necessarily need to follow one of those standards exactly.

Ideally, this mode would:

-  produce a positive image with low contrast and gentle highlight and shadow roll-off
-  avoid clipping individual RGB channels
-  avoid automatic creative grading
-  avoid strong saturation, vibrance or colour styling
-  avoid added sharpening, grain, glow or halation
-  preserve the widest practical colour gamut
-  export as a 16-bit TIFF with an embedded wide-gamut colour profile
-  produce consistent results across an entire roll when the scans use the same exposure and white balance
-  optionally preserve the film stock’s underlying colour relationships while correcting the orange mask and scanner-related colour errors

Why I want this

I would like to separate negative conversion from the final creative rendering.

My intended workflow would be:

1. Use NegPy to perform the technically difficult parts of the conversion, such as orange-mask removal, inversion, colour correction and tonal mapping.
2. Export a flat, neutral, high-information positive master.
3. Import that master into Lightroom.
4. Apply my own film-stock-specific presets and make per-image adjustments there.

For example, I may want to scan Fuji Natura 1600 and retain enough information to create a soft, pastel interpretation in Lightroom. For Kodak Gold, I may want to apply a warmer and more saturated rendering.

At the moment, a normal positive conversion may already include decisions about contrast, density, colour balance and photographic paper response. These decisions can make the initial result look good, but they may also reduce the flexibility available for later editing or make different film stocks converge toward a similar neutral rendering.

A flat-master mode would give users a clean and consistent starting point while still allowing the existing NegPy rendering workflow to remain available for users who want a finished image directly from the application.

Possible implementation

This might be implemented as a dedicated output mode or preset, for example:

* Flat Positive
* Neutral Master
* Digital Intermediate
* Log-style Positive

jboneng:
## Summary

Adds a new **output intent** that converts negatives into a flat, neutral, high-information
positive, a digital intermediate meant for further editing in Lightroom, Darktable,
Photoshop, etc. The goal is to maximize retained tonal/color information and minimize
creative interpretation, while staying consistent across a roll.

When the **"Flat — for editing elsewhere"** intent is enabled:

- A low-contrast, gentle-roll-off transfer curve replaces the H&D "print" curve
  (no hard clipping at either end).
- Creative/destructive stages are skipped: auto density/grade, cast removal, surround,
  flare, paper Dmin, retouch, lab (denoise/crosstalk/vibrance/CLAHE/sharpen/glow/halation),
  local dodge/burn, toning, and finish (vignette).
- Export defaults to a wide-gamut **16-bit master** (TIFF, with an optional Linear DNG)
  with an embedded ICC profile.
- For camera RAWs, the sensor-native linear RGB is mapped to **ProPhoto-linear** using the
  camera's `rgb_xyz_matrix` (row-normalized to preserve neutrals) so the wide gamut is
  colorimetrically meaningful rather than assumed.

## UX

- New **OUTPUT INTENT** section in the export sidebar with the flat toggle, a master format
  selector (16-bit TIFF / Linear DNG), and a **Preview flat** peek button.
- The H&D curve chart reflects the flat curve while peeking.
- Roll consistency: a hint + one-click **Bake roll baseline** button engages roll-locked
  normalization bounds so every frame on a roll is processed identically.
- The chosen label avoids jargon for non-technical users, "flat output" is a session
  preference that persists across restarts.

## Implementation notes

- `RenderIntent` enum (`PRINT` / `FLAT`) added to `ExposureConfig`, with serialization and
  legacy-default migration.
- `flat_master_config` / `flat_export_config` helpers derive the flat pipeline + export
  settings from the current config (mirrors the baseline-compare pattern).
- Flat renders are forced down the CPU path for deterministic parity.
- New `infrastructure/display/camera_color.py` handles the camera→ProPhoto matrix
  (export-time, so the normal render pipeline is untouched).

## Test plan

- [x] `uv run pytest` — 754 passed, 4 skipped, 1 xfailed.
  - All 20 flat-master tests pass (DNG roundtrip skips when optional `pidng` isn't installed).
  - Note: `test_export_presets.py::test_output_dir_subfolder_of_source` fails **on Windows only**
    and is **pre-existing / unrelated** (test hardcodes a `/` path while `os.path.join` yields
    `\` on Windows), passes on Linux/macOS CI.
- [x] New tests cover: flat curve is low-contrast & monotonic with no clipping; engine skips
  creative stages; config/serialization round-trips; flat routing; roll-lock gives identical
  output for identical patches; camera→ProPhoto matrix is neutral-preserving.
- [x] Manual: export a roll with flat intent, confirm 16-bit wide-gamut TIFF opens neutral
  and consistent across frames in an external editor.
  
  Fixes #278


marcinz606:
nice, I might reorder the export panel UI later on but let's merge it as it is :)

PR #299, opened 2026-06-25
Unify manual crop into a single draggable/resizable tool

reederphill:
## Summary
- Replaces the separate Manual-draw and Move crop tools with one Crop tool: drag corners to resize, drag inside to move, click outside an existing rect to draw a fresh one. Continuous drag updates render live and the tool stays open afterward (no one-shot draw-then-close).
- While the crop tool is open, the canvas previews the full uncropped frame (`PipelineContext.crop_preview_full`) instead of the already-cropped image, so you can resize the crop without losing context of what's outside it.

## Bugs found and fixed while building this
- Corner handles snapped/jumped while dragging — the inverse raw→screen lookup used a coarse quantized grid search; now exact for the 4 corner points.
- Aspect-ratio locking distorted the rect — ratio math was applied to raw normalized-space deltas, which only matches physical aspect ratio when the source image is square. Moved the ratio/min-size enforcement into screen-pixel space.
- The crop overlay drew the rect's fine-rotation-transformed corners as a tilted parallelogram, but the actual crop (`get_manual_rect_coords`/`CropProcessor`) only ever takes the axis-aligned bounding box of those corners — it's a plain array slice. The overlay now draws that bounding box so what you see matches what gets cropped.
- Dragging the crop rect caused a full CPU recompute of exposure/retouch/lab/local on every mouse-move, because the engine's base-stage cache key bundled the entire `GeometryConfig` (including crop-selection fields that don't affect pixel output while `crop_preview_full` bypasses the final crop). The cache key now excludes those fields in that mode.

## Test plan
- [x] `make all` (format, lint, type check, full test suite) passes
- [ ] Manually verify in the running app: drag corners to resize (with and without a locked aspect ratio), drag inside to move, draw a fresh crop from outside the rect, and adjust fine rotation while the crop tool is open

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -524,8 +524,13 @@ class AppController(QObject):
             self._handle_dust_pick(nx, ny)
 
     def set_active_tool(self, mode: ToolMode) -> None:
+        crop_tool_changed = ToolMode.CROP_MANUAL in (self.state.active_tool, mode)
         self.state.active_tool = mode
         self.tool_sync_requested.emit()
+        if crop_tool_changed:
+            # Entering/leaving the crop tool swaps between the full uncropped preview
+            # and the normal cropped preview, so the canvas must re-render immediately.
+            self.request_render()
 
     def cancel_active_tool(self) -> None:
         if self.state.active_tool != ToolMode.NONE:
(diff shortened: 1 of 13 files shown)


marcinz606:
I've rebased to current main (it was 5 commits behind) + did some small UX tweaks here: https://github.com/marcinz606/NegPy/pull/303

marcinz606:
thanks!

Issue #293, opened 2026-06-23
Feature Request: Support for pie-usb SANE backend / Reflecta ProScan 7200


hullrich:
I am trying to get my Scanner (Reflecta ProScan7200) working with NegPy (0.28.0).
Running NegPy from terminal, i get this log message: `Device ... has no recognised film sources — skipping`

For the Reflecta Scanner, SANE uses the [sane-pieusb](http://www.sane-project.org/man/sane-pieusb.5.html) Backend.
Scanning with X-Sane works fine, and i am able to scan a rudimentary image with a Test Script ([testscript.py](https://github.com/user-attachments/files/29247684/testscript.py)).
I am still figuring out how to change the default options though.

Looking at the code i noticed 2 issues with the sane backend:
1. NegPys [sane_backend](https://github.com/marcinz606/NegPy/blob/main/negpy/infrastructure/scanners/sane_backend.py) seems to check for a `source` option in [` _detect_caps`](https://github.com/marcinz606/NegPy/blob/main/negpy/infrastructure/scanners/sane_backend.py#L254C1-L268C86), which my scanner (and likely the pie-usb backend) does not provide.
2. The [IR dustremoval detection](https://github.com/marcinz606/NegPy/blob/main/negpy/infrastructure/scanners/sane_backend.py#L273) might cause issues, since it too searches for a option that is named "ir" or similar.
`sane-pieusb` named the option to apply IR dust removal `clean-image` (see details nr. 9, 19, 20).

**Would it be possible to adjust the `source` and `ir` checks so that pie-usb devices can be recognized correctly?**
And maybe allow some other options (8, 18, 21, 23, 24) to be set? (this could also be in the config to not bloat the UI)

I would be happy to help out with testing or providing more data if that's useful.
I am not very familiar with the low-level SANE internals though.

Full output from scanner.get_options() for Reflecta ProScan 7200:
<details>

```
(0, '', 'Number of options', 'Read-only option that specifies how many options a specific device supports.', 1, 0, 4, 4, None)
(1, None, 'Scan Mode', '', 5, 0, 4, 0, None)
(2, 'mode', 'Scan mode', 'Selects the scan mode (e.g., lineart, monochrome, or color).', 3, 0, 9, 5, ['Lineart', 'Halftone', 'Gray', 'Color', 'RGBI'])
[...]

PR #308, opened 2026-06-26
Fix/scan

hullrich:
See https://github.com/marcinz606/NegPy/issues/293#issuecomment-4803483758
- Force pieusb scanners into `Color` mode because `RGBI` is unreliable.
- Add hardware-specific flags for `pieusb` scanner option handling.
- Add an internal IR strategy for firmware-level dust removal, so the code handles that path cleanly.

This keeps pieusb behavior consistent and reduces edge cases around scanner mode and dust handling.

### DIFF
--- a/negpy/infrastructure/scanners/sane_backend.py
+++ b/negpy/infrastructure/scanners/sane_backend.py
@@ -352,16 +352,16 @@ class SaneBackend:
     def _set_pieusb_flags(self, dev, capture_ir) -> None:
         """Apply hardware-specific optimizations for pieusb scanners."""
         opts = {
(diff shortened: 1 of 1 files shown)


marcinz606:
that solves the issue without my changes in https://github.com/marcinz606/NegPy/tree/fix/scan branch?

hullrich:
No, these are your changes https://github.com/marcinz606/NegPy/tree/fix/scan + a fix for the IR option in there.
(forcing the `Color` option + backend flags as described in https://github.com/marcinz606/NegPy/issues/293#issuecomment-4803483758)

PR #364, opened 2026-06-30
fix(tutorial): keep tutorial overlay body text within popup width

### LINKED ISSUE #357: Tour text is slightly too wide for the view it is in
The text in the tour steps is slightly wider than the view it's in, so a few pixels are cut off on one side (and you can scroll it side-to-side). This is on macOS 26.5.1.

Image

seanharding:
QTextDocument.textWidth is the content width — Qt adds 2 * documentMargin around it, and the popup QFrame's 1px stylesheet border further trims the body viewport. The previous textWidth left the document a few pixels wider than the viewport, enabling tiny horizontal panning. Also enables WrapAtWordBoundaryOrAnywhere so long unbreakable strings (e.g. code paths) can't push the document wider than the viewport.

Fixes #357

### DIFF
--- a/negpy/desktop/view/widgets/tutorial_overlay.py
+++ b/negpy/desktop/view/widgets/tutorial_overlay.py
@@ -4,7 +4,7 @@ import sys
 from typing import TYPE_CHECKING, Callable, Optional
 
 from PyQt6.QtCore import QEvent, QObject, QRectF, Qt, pyqtSignal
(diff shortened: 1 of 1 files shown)


seanharding:
In my local testing, this seems to fix the issue. But full disclosure, I'm not very familiar with QT, so it's possible there's something I'm missing.

Issue #51, opened 2026-01-14
Can't open folders


thetalkingdrum:
Environment : Mac OS 14.8.3 

The process quits when trying to open a folder. Opening files works fine. 

From the terminal: 

[2026-01-14 10:01:14] INFO     negpy.src.infrastructure.loaders.native_picker: Native folder picker returned 48 files from /Volumes/KingstonSD512/Juan scans mega folder/Juan films mine scans Pakon/kodak 50D/chemvert
Numba workqueue threading layer is terminating: Concurrent access has been detected.

 - The workqueue threading layer is not threadsafe and may not be accessed concurrently by multiple threads. Concurrent access typically occurs through a nested parallel region launch or by calling Numba parallel=True functions from multiple Python threads.
 - Try using the TBB threading layer as an alternative, as it is, itself, threadsafe. Docs: https://numba.readthedocs.io/en/stable/user/threading-layer.html

zsh: abort      /Applications/NegPy.app/Contents/Resources/negpy/negpy

Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]

thetalkingdrum:
That would be great. I have not been in any great hurry to update to Tahoe. I think quite many Mac users are not on 26 yet as it has been a bit unpopular.  I just updated to Sequoia that I had planned to do anyway, but no dice. I might do a clean install to an external disk to see if I have some conflict in my system or if it really needs Tahoe.

thetalkingdrum:
I did a clean install of Sequoia on an external drive and tested NegPy there. Then I updated that install to Tahoe and tested again. In both cases, NegPy crashed when I clicked “Pick Folder,” so updating to Tahoe didn’t fix the issue in my case.

I’m on a 2021 16-inch M1 MacBook Pro.

What’s a little weird is that NegPy seems to work a bit better on the clean external install than on my original internal install. On the external drive, it can open some folders without crashing, but on my internal drive those same folders cause a crash.

Is there some kind of database or other files I could delete on my original install to completely reset NegPy?

Are there any logs I could send that would help you debug this? I could also send a folder that reliably causes the crash, since some folders seem more likely to trigger it than others.

thetalkingdrum:
Well, now I actually got it to load the the folder from the screenshot above. But a lot of the files are garbled and look like this one.  I was thinking of opening a separate issue on this but I wonder if it could be connected?


[...]

PR #426, opened 2026-07-09
Multi-core CPU rendering, HQ refresh fix, and true zoom % readout

linkmodo:
Performance and viewer fixes for the CPU rendering path, with a runtime failsafe for the multi-core kernels.

### Multi-core CPU rendering
The hot per-pixel Numba kernels ran single-threaded, so a CPU render pinned one core while the rest sat idle. The dominant kernels now run `prange` over the outer row loop:
- **Print curve** and **chemical toning** — each pixel is independent, so output matches serial.
- **CIELAB conversions** (`rgb_to_lab_working` / `lab_to_rgb_working`) and the **working-space OETF**, rewritten as `prange` kernels over an `(N, 3)` pixel list. Profiling showed the Lab stage (CLAHE + sharpening each do a full rgb→lab→rgb round-trip) was ~5.6s of an 8s full-res render in pure single-threaded numpy.

A full-resolution (24MP) CPU render drops from **~10s to ~2.7s** on a 24-thread machine (print curve alone ~17x in isolation). All characteristic-curve, toning, colour/lab/OETF, and GPU/chart parity golden tests pass; serial vs parallel output max diff is 2.3e-10 on a 24MP render.

### Crash failsafe for parallel kernels (based on macOS crash reports)
Crash reports on macOS during rendering/export trace to the pinned `workqueue` threading layer, which **hard-terminates the process when two threads invoke `parallel=True` kernels concurrently**. The render worker, export worker and the CPU display ICC LUT (parallel since before this branch) can all overlap; the multi-second kernels added here made that overlap routine. Three layers of defence, all via a new `parallel_njit` decorator (`kernel/system/parallel.py`) that every parallel kernel now uses — including the two pre-existing ICC LUT kernels:

1. **Root-cause fix** — parallel invocations are serialized behind a process-wide lock, so two threads can never be inside the workqueue scheduler at once. Each kernel already saturates all cores, so this costs nothing.
2. **True failsafe** — every kernel is compiled in **both serial and parallel variants**, dispatched per call on a runtime flag. The serial variant is the plain `@njit` path that has always been stable, so disabling provably reverts.
3. **Platform default** — parallel ON everywhere, **OFF on macOS** until the lock fix is verified on real hardware there. Users can override either way with `cpu_parallel = true/false` under `[performance]` in `override.toml`.

Implementation note: the serial variant is always compiled `cache=False` — numba's disk cache is keyed by source location, so the two variants of one kernel share a cache slot and load each other's binaries (a "serial" call silently executed the cached *parallel* object; verified in both directions). A regression test guards this, plus a 6-thread concurrency hammer.

**Asking for macOS testers**: confirm the default (serial) is stable when GPU rendering is toggled off; then set `cpu_parallel = true` to test whether the invocation lock alone fixes parallel mode so parallel can be enabled by default for macOS users in future iteration.

### HQ toggle showed the stale low-res render
`run_pipeline` built the engine cache key (`source_hash`) from the file hash plus flatfield/rgbscan/linear-raw tokens, but **not the buffer resolution**. Toggling HQ re-decodes the same file at full resolution with unchanged settings, so the key was identical, the engine cache reported "nothing changed", and the viewer kept the low-res render. Fixed by folding the buffer resolution into `source_hash`.

### True pixel zoom % + Fit / Original buttons
The zoom label showed the internal fit-relative `zoom_level` (1.0 = fit-to-window) as a percentage, so a fitted image always read "100%". Now:
- The label reports **true pixel zoom** (100% = 1 image pixel : 1 device pixel).
- Added **Fit to Window** and **1:1 (Original)** buttons next to the slider, wired to keys `0` / `1` / `2`. The 1:1 tooltip notes it shows a lower-resolution preview unless HQ is enabled.
- Fixed `fit_to_window()`, which set `zoom_level` to the fit scale even though the shader already fits at 1.0 — "fit" actually shrank the image.
- The % readout refreshes on viewport resize and on every new render buffer, so switching HQ/low-quality updates it immediately.

Full suite: 1225 passed (the one failing `test_export_presets` case is a pre-existing Windows path-separator assertion unrelated to this branch). Branch is rebased on current `main` (post-#425).


linkmodo:
Made enhancements to CPU rendering pipeline for both curves and color conversion. Prior they were single threaded, HQ rendering took over 10 seconds. Now they are almost as fast as HQ GPU preview rendering, and non-HQ rendering is almost instant. Fixed zoom level display calculation, added quick access button for fit-to-screen and 1:1, and some other minor UX fixes.

marcinz606:
I actually used `parallel=True` initially but that caused some instability/crashes on some setups, especially on macOS if I remember correctly

marcinz606:
But that was long time ago. I'll test a bit myself and also ask on https://github.com/marcinz606/NegPy/discussions for further testing

linkmodo:
Gotcha. I have a AMD Ryzen 9900X. Processing HQ preview image with GPU acceleration turned off took over 10 seconds, looked at resource monitor on core usage, and that's why I started tweaking. If causes instability maybe can consider a settings toggle to enable multi-core, or auto detection of processor type on startup to disable if Apple Silicon doesn't support it. Thank you.

marcinz606:
toggle/settings with safe defaults per platform sounds like good solution

linkmodo:
I'll see if I can implement some sort of system check to auto enable (then toggle not needed).

If not, adding a toggle next to GPU/CPU when CPU pipeline is enabled. But if user turn on multicore and crashes, then may need a safe mode to restart the app with default settings..

marcinz606:
override through `NegPy/override.toml` is also another safety option. The crashes/stability issues happened on preview render, mostly slider movement and sometimes on export so I think app should still start fine. Maybe keep multicore false by default (safe option) as whole CPU rendering path is more of a fallback

linkmodo:
Implemented a "failsafe" for "cpu failsafe" option. macOS defaults to serial processing if WebGPU is toggled off, but can force enable parallel in override.toml to harness the benefit of modern multi-core processors. On newer processors, the difference between GPU and CPU rendering speed is marginal, and maybe slightly power efficient.

thetalkingdrum:
Here is the original issue for reference, a bit long and messy, but the first post is probably the relevant part: https://github.com/marcinz606/NegPy/issues/51

linkmodo:
> Here is the original issue for reference, a bit long and messy, but the first post is probably the relevant part: https://github.com/marcinz606/NegPy/issues/51
> 
> 
> 
> 

Thank you for the reference! I think invoking concurrent parallel processing on workqueue was the root cause of crash during normal workflow.. Would you mind testing multi core rendering on Mac by forcefully enable new parallel processing pipeline in config file to test?

thetalkingdrum:
> Thank you for the reference! I think invoking concurrent parallel processing on workqueue was the root cause of crash during normal workflow.. Would you mind testing multi core rendering on Mac by forcefully enable new parallel processing pipeline in config file to test?

I did some quick tests right now, loading some folders and doing some quick edits. And if I did it right (uncomment cpu_parallel = true in override.toml, toggle gpu off with the lightning button in the GUI), so far so good. No crashes yet.

PR #431, opened 2026-07-09
Camera scanning: narrowband-RGB and white-light capture

light-sntchr:
Adds a **Camera Scanning** tab that captures negatives with a tethered camera and imports them straight into NegPy. Proposed and discussed in #429.

**Reworked onto gphoto2**, as you suggested. The proprietary Sony SDK is gone, and with it the external C++ helper, its build step and the licence question. The capture layer is now pure Python over LGPL libgphoto2, and it lost more code than it gained.

> ⚠️ **macOS and Linux only.** libgphoto2 has no Windows build, so there are no wheels and the tab cannot connect there. I have an old Windows machine and can dig into it separately — I'd rather not block on it.

Two modes, chosen automatically by what hardware is present:

- **Normal camera scanning** — one exposure under any light, imported as an ordinary RAW. No extra hardware.
- **Narrowband RGB** — with jackw01's Scanlight connected, the light flashes R → G → B while the camera captures each exposure, and the three RAWs go straight to the existing RGB Scan merge (#288). One LED at a time means no cross-channel contamination.

Plus per-channel ETTR auto-calibration on the film base (with a raw-Bayer clip guard on the highlight that becomes the black point after inversion), film-stock presets, live view with the hardware focus magnifier — click to magnify, click again for the full frame — and per-roll auto-numbered output that imports and merges after each shot. Cameras are found on the USB bus; there is no address to type, no login, no pairing.

### What this does and doesn't touch

- **Your merge and inversion are untouched.** The capture layer only *feeds* `features/rgbscan`.
- Existing source files changed: `controller.py`, `right_panel.py`, `pyproject.toml`, `build.py`. Plus an architecture section in `CLAUDE.md` and a new `docs/CAMERA_SCANNING.md`. Everything else is new, self-contained modules under `infrastructure/capture/` and `services/capture/`.
- `gphoto2` is an **optional dependency** (`--group camera`, next to your existing `scanner` group). Without it the tab shows a one-line setup hint and nothing else in NegPy notices. `pyserial` (for the light) is the only addition to the main deps — happy to move that too if you'd rather.
- **Packaging:** `build.py` re-adds libgphoto2's `camlibs`/`iolibs` trees by hand. `--collect-all` flattens them in among the other binaries, but the library `dlopen`s its drivers from exactly that directory layout — without it a packaged app loads libgphoto2 and then fails to find *any* camera. Verified: the running `.app` has `camlibs/ptp2.so` open from inside the bundle. Mirrors your `scanner` group's "run `uv sync --group camera` before building" pattern.

### On more camera brands

**Only Sony is tested** — that is the hardware I own, and I won't claim more than I can show. But I went through `camlibs/ptp2/config.c` hunting for vendor assumptions, and closed the four I found:

- **Control names are looked up, not assumed.** `iso` and `shutterspeed` are the same everywhere; the aperture is not — `f-number` is the generic PTP name (Sony, Panasonic), while Canon, Nikon, Fujifilm, Olympus and Sigma expose `aperture`. No body offers both.
- **The RAW suffix comes from the camera.** It used to hardcode `.ARW`, so a Canon would have written a `.CR3` under a Sony name.
- **The still is taken into memory.** Sony already does; Canon and Nikon default to the card and won't shoot without one. The option is labelled differently on each (`sdram`, `Internal RAM`), so NegPy matches the word.
- **The focus magnifier is a per-vendor table** rather than a Sony constant. Sony packs ratio and position into one property, so a click magnifies *and* aims; Canon (`eoszoom`) and Nikon (`liveviewimagezoomratio`) split them and their coordinate space is unknown to me, so a click magnifies where the body already looks. Bodies with no magnifier disable the feature. Previously the missing property raised straight through a Qt slot, which aborts the process.

Two more robustness fixes fell out of that: a value the body doesn't offer (a shutter label from the fallback ladder) now warns instead of aborting a scan, and I dropped the white-balance control entirely — NegPy decodes with a fixed neutral WB, so the camera's setting never reached a scan anyway.

libgphoto2 lists **205 Canon and 70 Nikon bodies with live view**, so the door is genuinely open — but someone with that hardware has to walk through it. Worth noting: my a7C II isn't even in libgphoto2's camera list; it binds to the generic `USB PTP Class Camera` entry and the ptp2 driver applies the vendor extensions anyway.

### Testing

- Full suite green: **1346 tests**, `make all` clean. Every capture test runs against a fake libgphoto2, so **CI needs no hardware**.
- Verified end-to-end on real hardware (a7C II): live view, focus magnifier, calibration, triplet scan → merged + inverted, and unplugging the camera mid-stream.
- **Only macOS is tested.** libgphoto2 is a Linux-first project so Linux should be at least as good, but I can't vouch for it.

### Four things libgphoto2 does that will bite anyone who touches this

Each has a guard and a comment in `gphoto.py`; I found all four on hardware, none of them in the docs.

1. Reading the value of a **choice widget with no choices** dereferences NULL and kills the process with SIGSEGV — no `except` can catch it. That is `f-number` on a lens with no electronic diaphragm, i.e. most enlarging and macro glass.
2. **Property writes are asynchronous** (~1-2 s). A read-back straight after a write returns the old value.
3. After a still, the **event queue must be drained** or the *next* capture fails with a bare `[-1] unspecified error`.
4. Choice strings run through **gettext**, so `focusmode` reads `Manuell` on a German desktop. Only the message locale is pinned — touching `LC_ALL` would change Qt's number formatting.


marcinz606:
thanks! i will check it out during the weekend

marcinz606:
https://github.com/jim-easterbrook/python-gphoto2
https://pypi.org/project/gphoto2/
can you see if you can get it to work with this? maybe we can make it vendor-agnostic and not require propertiary sdks

light-sntchr:
Gave it a quick test on my a7C II — works well. Live view, shutter control, capture to host, and even the focus magnifier including x,y positioning, all through gphoto2. Three shots back-to-back took 5.1 s, a touch faster than the Sony SDK path.

Curious detail: the a7C II isn't in libgphoto2's camera list at all. It binds to the generic `USB PTP Class Camera` entry and the ptp2 driver applies Sony's vendor extensions anyway.

So I'll have a go at reworking this onto gphoto2 — that drops the proprietary SDK and the C++ helper entirely. One thing to know up front: python-gphoto2 only ships Linux and macOS wheels (libgphoto2 has no Windows support), so camera scanning would be macOS/Linux for now. I'll look into Windows separately.

marcinz606:
great! SANE scanning is also macOS & Linux only for now. Let's treat it as incentive to move people from Windows to Linux/macOS for now :)

linkmodo:
> great! SANE scanning is also macOS & Linux only for now. Let's treat it as incentive to move people from Windows to Linux/macOS for now :)

Windows user here. Till I can afford a mac X-D

rohanpandula:
Hey Team, just a quick note here /sharing why I went with the official SDK over gphoto2. I did build and test a clean-room gphoto2 wrapper for my own a7CR rig first.

Two gaps: Pixel Shift Multi-Shooting isn't in gphoto2 yet (haven't even confirmed its capture loop can pull multiple RAWs off one release), and Wi-Fi capture is blocked upstream, Sony's 0x9209 property call doesn't complete over ptpip, confirmed independent of my code. Neither's gphoto2's fault, Sony's PTP/IP side is just thin.

So maybe this isn't either/or: libgphoto2 as the default agnostic backend, vendor SDKs opt-in for whoever needs PSMS, Wi-Fi, or whatever else comes up per-brand. Same helper contract underneath either way.

Can share my PTP-opcode notes if useful some of what Robin ran into today (generic "USB PTP Class Camera" binding, ptp2 applying Sony's vendor extensions) I'd already mapped. 

Edit here are my notes: https://gist.github.com/rohanpandula/995d1c8933f43bbf4837a599205bbc34

linkmodo:
Suggesting an alternative workflow for Sony: Where instead of live-capture of the camera scan, enable SFTP to just wirelessly transfer files to the "Hot Folder" directory setup in NegPy.

No SDK integration, no additional coding, and images are instantly available and the image capture can happen remotely.

light-sntchr:
@marcinz606 reworked onto gphoto2 as you suggested: the proprietary Sony SDK and the C++ helper are both gone, capture is now pure Python over libgphoto2 (USB, macOS/Linux). It's vendor-agnostic. Every camera control is looked up rather than hardcoded, so it should work with any body libgphoto2 supports, though only Sony is tested. Full detail in the PR description. This is ready to go as the baseline; Wi-Fi and Pixel Shift will follow as a separate opt-in PR on top.

@rohanpandula thanks, and +1. The SDK-as-opt-in backend is clearly the right call for the two things libgphoto2 can't do. The `Camera` protocol in this PR is already that seam, so once this lands I'll take it on as a follow-up: Wi-Fi + Pixel Shift over the Sony SDK (Sony-only, opt-in), building on your `sony-capture-psms-multiframe` work. Your Wi-Fi/`0x9209` notes save me a ton of digging there, much appreciated.

@linkmodo low-friction, and it mostly works today: point the camera's FTP at a folder and NegPy already groups the R/G/B files into triplets. The catch for narrowband RGB is that it drops live view and focus aids, and NegPy can't fire the Scanlight red→green→blue in lockstep with each exposure or run the auto-calibration, so you'd be hand-coordinating the light and shutter per channel. Great for shoot-and-import; the coordinated scan needs the tether.

marcinz606:
thanks! I'll test it out over the weekend and likely merge :)

rohanpandula:
I finished a software-complete safety/integration pass for this PR and opened a stacked update here: https://github.com/light-sntchr/NegPy/pull/1

Because `camera-scanning` is owned by @light-sntchr and I do not have push access to that fork, merging the stacked PR into `light-sntchr:camera-scanning` is the clean way to update this original PR.

The pass fixes the failure modes I could reproduce without hardware: mixed old/new files after a failed retake, duplicate/stale same-path imports, captures dropped during concurrent discovery, unsafe calibration successes (zero/non-finite signal, rejected shutter writes, unresolved clipping), X-Trans half-size aliasing, capture process-mode handoff, cancellation UI dead-ends, leaked calibration RAWs, unsafe roll paths, processed-JPEG capture, and camera-control disconnect exceptions.

Fresh verification on the exact published tree:

- 1,483 passed, 4 skipped, 7 deselected
- Ruff check/format and `ty` clean
- public Fuji X-Trans RAF triplet decoded/merged to finite output
- macOS DMG built; bundled app launched; libgphoto2 camlibs/iolibs present in the app bundle

No physical camera or Scanlight was attached for this final pass, so the original a7C II run remains the hardware evidence. The stacked PR is currently draft and mergeable; it has the full fix/verification breakdown.

marcinz606:
merged, thanks!

PR #453, opened 2026-07-11
Use ZIP (Deflate) instead of LZW for TIFF exports (#238)

### LINKED ISSUE #238: Exported TIFF often ends up significantly larger than the original
I noticed that NegPy exported TIFFs often end up quite a bit larger than my original linear TIFFs, so I ran a few tests.

Results (same image):

Original TIFF (no compression): 128.6 MiB
NegPy LZW: 163 MiB
ImageMagick LZW: 140 MiB
ImageMagick ZIP: 108 MiB
Photoshop LZW: 161 MiB
Photoshop ZIP: 121 MiB

It seems like LZW performs quite poorly on this type of 16-bit scan data compared to ZIP compression, or even no compression, with file sizes growing larger than the original uncompressed tiff. 

I also noticed that ZIP results vary quite a bit between tools, which is likely related to different TIFF predictor settings (horizontal differencing) and internal encoding/layout choices.

Maybe it would make sense to switch the default TIFF export to ZIP (with the most efficient predictor settings)?

(Support for lossless JEPG XL might also be a cool feature down the road, as it compresses quite a bit better still..)

RP2:
Fixes #238 -- 16-bit TIFF exports are needlessly large because LZW compression performs poorly on photographic 16-bit data, often barely beating uncompressed.

## Root cause

`tifffile.imwrite` in the export and scan-capture codepaths hardcoded `compression="lzw"`. LZW is a byte-oriented algorithm designed for 8-bit palette images -- it does not exploit horizontal redundancy in 16-bit continuous-tone data, so 16-bit RGB TIFFs compress poorly.

## Fix

Replace `compression="lzw"` with `compression="zlib"` + `predictor=True` in all three TIFF-writing call sites. `predictor=True` enables horizontal differencing (TIFF Predictor 2), which stores pixel-to-pixel deltas instead of absolute values. Combined with Adobe Deflate (ZIP), this yields ~20-35% smaller files for 16-bit photographic content.

Both `tifffile` and `imagecodecs` are already project dependencies -- no new imports needed.

**Files changed:**
- `negpy/services/rendering/image_processor.py` -- export TIFF
- `negpy/services/scanning/writer.py` -- scan capture RGB + IR sidecar
- `tests/test_export_service.py` -- test helper consistency
- `tests/metadata/test_writer.py` -- test helper consistency

## Verification

- Existing round-trip tests pass (all 13 tests green)
- Written TIFFs carry the correct `Compression` tag (8 = Adobe Deflate) and `Predictor` tag (2 = horizontal differencing)
- Lint and type checks clean

### DIFF
--- a/negpy/services/rendering/image_processor.py
+++ b/negpy/services/rendering/image_processor.py
@@ -361,7 +361,8 @@ class ImageProcessor:
                 img_out,
                 photometric="rgb" if img_out.ndim == 3 else "minisblack",
                 iccprofile=icc_bytes,
-                compression="lzw",
+                compression="zlib",
+                predictor=True,
             )
             return output_buf.getvalue(), "tiff"
         elif fmt == ExportFormat.DNG:
--- a/negpy/services/scanning/writer.py
+++ b/negpy/services/scanning/writer.py
@@ -37,7 +37,7 @@ def write_tiff_16bit(result: ScanResult, path: str) -> str:
     fd, tmp_path = tempfile.mkstemp(suffix=".tif", dir=os.path.dirname(path) or ".")
     os.close(fd)
     try:
-        tifffile.imwrite(tmp_path, rgb, photometric="rgb", compression="lzw")
+        tifffile.imwrite(tmp_path, rgb, photometric="rgb", compression="zlib", predictor=True)
         os.replace(tmp_path, path)
     except Exception:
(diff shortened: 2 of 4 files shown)


thetalkingdrum:
Yes please. LZW actually made the tiffs larger than plain uncompressed.

PR #460, opened 2026-07-12
Clockwise-positive fine rotation + Straighten reference-line tool

linkmodo:
## Summary

Two alignment-workflow improvements that share one sign convention, plus a small dialog polish.

### Clockwise-positive fine rotation

The Fine Rotation slider now reads in the photographer's convention: **positive turns the image clockwise on screen, negative counter-clockwise — under every flip state**. The crop tool's rotation-handle angle badge and the `Alt+R` / `Alt+Shift+R` shortcut descriptions follow the same convention.

The sign flips at the **display boundary only**; `geometry.fine_rotation` keeps its internal cv2/warp convention, so **every previously saved edit renders exactly as before** (negating the stored convention would have silently mirrored the tilt of existing edits, with no clean migration path). Because the pipeline applies flips *before* fine rotation, the stored angle's on-screen direction is flip-independent — verified empirically by rotating marker pixels with and without pre-applied flips — so the single display negation holds in all flip/rotation states.

### Straighten reference-line tool

A ruler button under **Geometry → Alignment** (default key **L**, rebindable in Customize, listed in the Keyboard Shortcuts dialog under Geometry): activate, draw a line along a tilted horizon or a leaning vertical edge, and the image rotates to match on release — one-shot, like Lightroom's straighten line.

- **Intent detection**: lines closer to horizontal level to the horizon; lines closer to vertical plumb (e.g. a building edge). Drawing the same line from either end gives the identical correction.
- A live badge previews the correction while dragging: *"Level +1.43°"* / *"Plumb −5.71°"*, in the CW-positive display convention.
- The delta is measured in display space and **added to the current fine rotation** (clamped to ±45°), landing in the same field the slider and crop handles drive — so it composes correctly under flips and 90° turns, and remains adjustable afterwards.
- Guards: sub-8px drags ignored (accidental clicks), right-click never draws, Esc cancels an in-progress line; screen-space measurement makes zoom/pan/HQ irrelevant to accuracy.

### Dialog polish

The Keyboard Shortcuts dialog's Customize/Close buttons set widget-level stylesheets with hardcoded backgrounds, which beat the app stylesheet in every state and killed the global hover-grey / pressed-red feedback. Customize now inherits the standard button states; Close uses the app's `primary` treatment (accent fill with its own hover/pressed shades).

## Testing

Verified in-app on Windows: slider direction under H/V flips and 90° rotations, straighten on horizons and verticals (both lean directions), badge signs matching the slider, `L` toggle + rebinding, saved edits unchanged. Unit tests cover the straighten math (axis snapping, both lean directions, direction-agnosticism, exact-axis no-op, ±45° bound) and a headless end-to-end drives press/drag/release through the overlay asserting emitted deltas. 103 geometry + shortcut-registry tests pass; ruff clean.


marcinz606:
that straighten-to-line is great!

PR #473, opened 2026-07-12
Camera scanning: lock each preset's exposure, plus live-view/calibration fixes

light-sntchr:
Follow-up polish to the camera-scanning tab from #431, from testing a packaged
build on a Sony a7C II + RGB Scanlight.

## Fixes
- **Debounced live-view setting writes** so stepping ISO/shutter/aperture quickly
  no longer queues a ~1-2 s verified write per intermediate step.
- **ETTR calibration hardened + sped up** on a marginal base: a badness-scoring
  shutter search and an iterating clip-down place every channel just under
  clipping instead of failing.
- **Focus magnifier is fire-and-forget**, so click-to-zoom no longer freezes the
  live-view preview while the body engages.
- **ISO stepper offers only fixed single-shot ISOs** (drops Auto and the
  Multi-Frame-NR pseudo-ISOs a scan can't use); the body is nudged off Auto.
- Scan-tab polish: the "connect the camera" hint hides once connected, the
  status LED reads "Camera", and the sliders reflect the *preset*, not the live LED.

## Preset exposure locking (new)
A calibrated preset now stores the **ISO and aperture** the base was metered at,
and a scan **re-asserts the full exposure on the body before every frame**, so a
bumped dial can't silently falsify a scan. White-light and normal captures leave
the exposure free.

- A selected preset is shown **read-only** (sliders + exposure steppers disabled).
- The calibration pop-up gained ISO/aperture steppers to set before metering.
- **"Create a manual preset"** unlocks the sliders + steppers to build one by hand
  (steps through the body's own choices; greyed out with no camera).

Camera-agnostic: property names resolve per body (aperture is `f-number` on Sony,
`aperture` on Canon/Nikon/etc.). No vendor SDK is bundled or linked; `gphoto2`
stays optional. `make all` green with and without the camera group.


docs/CHANGELOG.md
```diff
@@ -14,7 +14,7 @@
...
+- New: **Camera Scanning** — a new tab captures negatives with a tethered camera and feeds the RGB Scan merge. Two auto-selected modes: **Narrowband RGB** (jackw01's open-source Scanlight flashes red, then green, then blue while the camera captures each exposure) and **normal** (a single exposure under any light, imported as an ordinary RAW). Includes per-channel ETTR auto-calibration metered on the clear film base, film-stock presets, and a live view with the camera's hardware focus magnifier and live ISO/shutter controls. Each preset stores and re-asserts the full exposure (ISO, shutter and aperture) on the body before every frame so a bumped dial can't falsify a scan, and can be built by hand as well as by calibration. Cameras are detected on the USB bus automatically — no address, login or pairing. macOS and Linux only (verified on Sony bodies, other brands need testing). @light-sntchr
```
marcinz606: 0.36.0 shipped :) you can skip updating the changelog, I'll do changelog anyway before next packaged release

light-sntchr: Ah got it, thanks! Dropped the changelog change.

PR #479, opened 2026-07-13
fix(capture): release the camera PTP session when the scan/preset windows close

bbatha:
## Summary
- The tethered gphoto2 camera session was only ever closed on a capture error or full app shutdown, so exiting the Live View window or the new-preset calibration pop-up left it held open indefinitely.
- Some bodies (Fuji in particular) get stuck in a tethered-capture state on the camera side until the session is cleanly exited — leaving it open past the last window that uses it makes the *next* connection attempt hang instead of reconnecting.
- The session is now released once neither camera window is open and nothing is mid-capture/calibration, with a suppression guard so switching directly between Live View and the calibration pop-up doesn't trigger a spurious close+reconnect.

## Test plan
- [x] `uv run pytest tests/test_capture_gphoto.py tests/test_capture_worker.py tests/test_capture_controller.py tests/test_scanlight_sidebar.py` — 100 passed
- [x] `uv run ruff check` on the touched files — clean
- [x] `uv run ty check` on the touched files — no new diagnostics (7 pre-existing, unrelated to this change, confirmed identical on `main`)
- [x] Verified against a real Fuji X-T5 over USB: repeated open → close → reopen cycles are fast (~0.6s) and reliable, with no hangs
- [ ] Could not force-reproduce the original multi-second hang on the test rig to empirically prove the exact failure mode; this fix closes the identified gap (session never released on dialog close) that is the most likely root cause

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -173,6 +173,7 @@ class AppController(QObject):
     capture_status = pyqtSignal(str)
     live_view_requested = pyqtSignal(LiveViewRequest)
     live_view_stop_requested = pyqtSignal()
+    camera_session_close_requested = pyqtSignal()
     live_view_focus_magnifier_requested = pyqtSignal(bool)
     live_view_focus_magnifier_pos_requested = pyqtSignal(int, int)
     live_view_camera_setting_requested = pyqtSignal(str, int)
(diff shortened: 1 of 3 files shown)


marcinz606:
don't have fuji so can't verify but looks reasonable :)

PR #489, opened 2026-07-14
Add roll-aware Auto Crop All and consensus detection

### LINKED ISSUE #476: Add roll-aware Auto Crop All before Batch Analysis
## Problem

Auto Crop can produce inconsistent frame boundaries across a roll of camera-scanned negatives. Dark scenes, bright light-table borders, ragged rebate, holder edges, and scanner streaks can cause an image-by-image detector to select scene detail or a canvas separator instead of the film frame.

Frames from the same roll usually share width, placement, top offset, and rotation. Confident detections can therefore calibrate ambiguous ones. However, running this calibration during export is too late: NegPy’s Batch Analysis expects images to be cropped first so sprocket holes and empty space do not contaminate the roll-wide exposure and color baseline.

## Proposed UX: Auto Crop All

Add a dedicated **Auto Crop All** action in the Crop panel, alongside the existing single-image **Auto** control.

- **Auto** continues to crop only the current frame.
- **Auto Crop All** analyzes all currently visible frames together before Batch Analysis.
- It runs in the background with progress and cancellation.
- Existing manual crops are preserved.
- Unsupported or ambiguous frames remain unchanged for manual review.
- Confident results are saved as each image’s explicit manual crop plus fine rotation, so all existing preview, Batch Analysis, export, and sidecar paths see the image as already cropped.
- The current implementation targets **Image only** mode and the current landscape workflow.

This makes roll calibration an explicit preprocessing step instead of hidden export behavior.

### 1. Collect crop evidence

Run detection on NegPy’s in-memory float32 preview after coarse rotation, flips, flat-field correction, fine rotation, and distortion correction.

For each frame, return a candidate containing:

- exclusive `(top, bottom, left, right)` coordinates
- deskew angle and confidence
- detected sides/corners and primary mask evidence
- geometry plausibility
- vertical edge profile for fallback placement

Detection remains local and deterministic: downsample the image, estimate rotation from coherent edge profiles, classify dark rebate and bright light table with hysteresis, and use low-texture gradients as fallback evidence. Require evidence on multiple sides to protect dark scenes and already-cropped images.

### 2. Build a roll template

Use trustworthy two-corner candidates to calculate:

- median frame width
- upper-percentile fallback width
- horizontal center and top offset
- median deskew angle
- median-absolute-deviation tolerances for rejecting false width and angle clusters

### 3. Resolve weak frames

In a second pass:

- expand small width shortfalls toward the calibrated median
- preserve a strongly supported corner and infer the opposite edge
- use nearby edge-profile evidence for severe failures
- replace isolated or weak angle estimates with the roll median
- leave frames with no film-edge evidence unchanged

### 4. Keep a safety border

Retain a small outward margin by default—currently 1% of the short canvas dimension. Padding must use the same pixel amount on all four sides; reduce the common amount when a canvas edge lacks room. The existing Crop Offset control can then move the crop farther inward when desired.

### 5. Persist safely

Convert successful results into normalized `manual_crop_rect` coordinates and add the detected deskew to `fine_rotation`. Before saving, reload the latest per-file settings and skip a result if the user added a manual crop or changed detection-affecting geometry while analysis was running.

Per-asset RGB triplet paths must be resolved from each asset rather than inherited from the active frame. Fresh assets should use NegPy’s normal clean, sticky-aware defaults, not the active frame’s creative edits.

Prevent Batch Analysis, export, contact-sheet generation, discovery, and thumbnail batches from starting concurrently with Auto Crop All. Cancel the worker during application shutdown.

## Likely integration points

- `negpy/features/geometry/batch_autocrop.py`: candidate detection, roll templates, and result conversion
- `negpy/desktop/workers/render.py`: background Auto Crop All worker
- `negpy/desktop/controller.py`: orchestration, persistence, conflict protection, and batch ownership
- `negpy/desktop/view/sidebar/geometry.py`: Auto Crop All button
- `negpy/desktop/session.py`: clean configuration for fresh assets

## Prototype status

A working prototype has been developed in the `EItanya/NegPy` fork on a local `feature/batch-autocrop` branch. It has been exercised locally against camera-scanned negatives and is producing substantially more consistent crops than independent per-image detection.

rohanpandula:
## Summary

- replace single-winner film-bound detection with confidence-scored, adaptive multi-threshold consensus while retaining the existing masks as corroborating evidence
- add **Auto Crop All**, a two-pass roll workflow that groups frames by target ratio, calibrates weak crops from trusted frames, applies roll-median deskew safely, and adds a uniform outward safety border
- process visible landscape frames off the UI thread with progress and generation-scoped cancellation; preserve manual, Film-edge, portrait, ambiguous, changed, or failed frames
- match normal preview inputs for camera WB, RGB triplets, flat-field correction, orientation, flips, fine rotation, and distortion; serialize heavy batch jobs and use a private preview cache
- persist explicit manual crop rectangles and fine rotation with dispatch-time conflict checks, atomic active-frame persistence, documentation, and regression coverage

## What this adds beyond #476

Issue #476 defines the **Auto Crop All** product workflow: visible-frame scope, landscape/Image-only guardrails, manual-crop preservation, roll calibration, progress and cancellation, conflict checks, RGB handling, and batch serialization.

On top of that workflow, this PR strengthens the underlying crop system:

- **Single-frame detector overhaul:** adaptive seven-threshold consensus replaces the old single best contour. Dark film on a light bed and bright film in a dark holder are handled explicitly; legacy inverse-threshold, black-hat, and edge masks only corroborate a result.
- **Confidence and abstention model:** detections expose supported sides and corners, geometry confidence, evidence sources, residual deskew, and absolute vertical-edge contrast. Low-signal, noisy, and unsupported frames abstain instead of receiving a plausible-looking crop.
- **Ratio-isolated robust calibration:** roll templates are built independently per target ratio with median/MAD outlier rejection, preventing mixed 3:2, 4:3, and other settings from contaminating one another.
- **Coordinate-safe roll deskew:** when weak angles are replaced by the roll median, crop bounds are transformed into the new rotation space and quantized outward before persistence, so the correction cannot clip film-edge pixels.
- **Pipeline and state safety:** detection matches preview decoding (camera WB/Linear RAW, RGB triplets, flat field, orientation and flips, existing fine rotation, and lens distortion), uses a private preview cache, fingerprints geometry-changing inputs, and persists the active frame only after storage succeeds.
- **Regression hardening:** tests cover low-signal noise, mixed ratios, pre-start and final cancellation races, stale results, persistence failures, and half-open rotation-bound enclosure.

## Scope

This first version intentionally supports **Image only** mode and landscape frames. Existing manual crops are never overwritten. Cancelling before completion persists nothing.

The implementation was developed independently from the behavior described in #476. The referenced local `feature/batch-autocrop` prototype was not published, available, or reused.

Closes #476.

## Verification

- `make test` — 1898 passed, 4 skipped, 7 deselected
- `uv run ruff check .`
- `make type`
- `uv run ruff format --check .`
- GitHub **CI** and **Metrics** checks
- three independent review passes, including low-signal noise Monte Carlo and rotation-coordinate enclosure repros


marcinz606:
works really nicely!

Issue #452, opened 2026-07-11
Exporting results in an error "failed to load file: 'ascii' codec can't encode character '\xd7' in position 27: ordinal not in range(128)"


theoryee:
Exporting results in an error "failed to load file: 'ascii' codec can't encode character '\xd7' in position 27: ordinal not in range(128)" and no resulting exported file.

This is running on a Windows 11
On a .NEF source file (is there a list of supported file types?)

theoryee:
I'm still seeing this on exporting jpg, tiff, png.  

Seems to work fine for dng, .jxl, webp now.

PR #490, opened 2026-07-14
Fix export crash from non-ASCII film format in EXIF metadata (#487)

### LINKED ISSUE #487: Exporting results in an error "failed to load file: 'ascii' codec can't encode character '\xd7' in position 27: ordinal not in range(128)"
I'm still seeing this on exporting jpg, tiff, png.

Seems to work fine for .DNG, .jxl, webp now.

RP2:
Fixes #487 -- the same `UnicodeEncodeError` still happened on JPG, TIFF, and PNG exports after #452 was merged. I only fixed half the problem.

## Root cause

The #452 fix covered strings read from source EXIF via `_decode_ascii`. What I missed is NegPy writes its own non-ASCII strings. The `FilmFormat` enum uses `×` (0xd7, multiplication sign) in `"4×5"` and `"8×10"` at `gear_models.py:15-16`. When a user has one of those formats, it flows through `build_metadata_payload` > `_build_custom_exif` > UserComment construction at `writer.py:128`, which had a bare `.encode("ascii")` and crashed.

There's a second crash site too; `_fold_user_comment_into_description` at `writer.py:633` used `raw[8:].decode("ascii", "replace")` to read the UserComment back out. That produces U+FFFD (replacement character) for any byte it can't decode, and U+FFFD isn't ASCII. When it hit `tifffile.imwrite(description=...)`, same crash.

DNG, JXL, and WebP weren't affected because `export.py:139-143` skips metadata for those three. That's why the reporter only saw it on JPG, TIFF, and PNG.

## Fix

Two lines in `negpy/features/metadata/writer.py`:

1. Line 128 -- `.encode("ascii")` > `.encode("ascii", "replace")`, same pattern `_exif_ascii` already uses at line 79.

2. Line 633 -- swapped the raw decode for `_decode_ascii(raw[8:])` then `.strip()`. That helper already round-trips through `encode("ascii", "replace")` so the output's guaranteed pure ASCII.

**Files changed:**
- `negpy/features/metadata/writer.py` -- both crash sites
- `tests/metadata/test_writer.py` -- 2 new regression tests

## Verification

- 18/18 pass in tests/metadata/test_writer.py (new tests cover FilmFormat crash and UserComment fold with non-ASCII bytes)
- 4/5 pass in tests/test_metadata_writer.py (1 skipped, needs exiftool)
- End-to-end simulation with `format="4×5"` on TIFF, JPEG, and PNG -- no crash
- Lint and format clean

### DIFF
--- a/negpy/features/metadata/writer.py
+++ b/negpy/features/metadata/writer.py
@@ -125,7 +125,7 @@ def _build_custom_exif(payload: MetadataPayload) -> dict:
 
     if user_comment_parts:
         lines = [f"{k.replace('_', ' ').title()}: {v}" for k, v in user_comment_parts.items()]
-        uc_bytes = b"ASCII\x00\x00\x00" + "\n".join(lines).encode("ascii")
+        uc_bytes = b"ASCII\x00\x00\x00" + "\n".join(lines).encode("ascii", "replace")
         exif[piexif.ExifIFD.UserComment] = uc_bytes
 
     if flags.exposure and payload.capture_exposure:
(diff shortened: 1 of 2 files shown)


RP2:
Small followup -- the ? in "Format: 4?5" maybe isn't ideal. A payload.film_format.replace("×", "x") before the encode would just emit "4x5". Not blocking, but figured I'd mention it.

PR #497, opened 2026-07-14
Add Nikon LS-5000 full-roll preview and RGBI capture

rohanpandula:
This PR adds a Nikon LS-5000 roll-scanning workflow to NegPy. It covers low-resolution roll preview, frame alignment, selected-frame capture, scanner-linear RGB and IR masters, safe stopping and retry, and post-import IR dust repair.

Testing used an LS-5000 ED with firmware 1.03 and an SA-21 fitted with the solderless SA-30 conversion.

## Roll preview and selection

The Scan panel can read all 40 addressable roll slots at 97 dpi and display them as a contact sheet. Users can:

- Select individual frames or ranges.
- Adjust each frame with a numeric Film Spacing Offset and slider. `Alt+Left` and `Alt+Right` make one-step changes.
- Reload a thumbnail after changing its spacing.
- Eject the film from NegPy.
- See scanner status for connection, preview loading, frame progress, stopping, completion, and recovery.
- Switch between an inverted positive preview and a non-inverted negative preview without contacting the scanner.
- Adjust the Metering crop from 0% to 30% per edge with a slider. This recalculates thumbnail brightness from saved preview data. It does not crop the negative, rerun scanner exposure, or move the film.

Roll thumbnails and final scans have separate quality displays. Preview is fixed at 97 dpi. Selected roll frames are captured at 4000 dpi and saved as 16-bit scanner-linear masters.

An inferred frame position, most often slot 1 behind a long clear leader, requires explicit review. NegPy binds that approval to the thumbnail, offset, scanner origin, and reviewed roll fingerprint. The capture worker reads a fresh roll index before the first fine scan and refuses a different, reordered, or stale roll.

Synthetic LS-5000 40-slot preview showing the Metering crop slider

The screenshot uses synthetic thumbnails. It contains no personal photographs and did not require a scanner session.

## Capture and output

Selected C-41 frames share one direct scanner session and one reservation. NegPy resolves the selected frame positions before capture, scans each frame, verifies and publishes its output, removes the large packed scratch file, and then acknowledges that the scanner may continue. A clean finish or safe stop sends one release at the end.

Each color frame produces:

- A 4000 dpi, 16-bit RGB TIFF master.
- One aligned 16-bit IR TIFF.
- An IR validity mask.
- A scan receipt with the physical slot, Film Spacing Offset, reviewed roll identity, clipping telemetry, focus-detail telemetry, and transport-smear result.

The packed scanner stream contains four transferred RGB sample groups and one aligned IR plane. NegPy averages the four RGB groups. The wire data does not prove whether the scanner combines multiple IR acquisitions in firmware, so the receipt records the IR multisample semantics as unresolved.

IR dust repair runs after import and reports whether it was applied, skipped, unavailable, or still pending. The scanner-linear RGB and IR masters remain unchanged.

Conventional silver black-and-white film uses the SANE route at 4000 dpi, 16-bit RGB, and 4x sampling. IR capture and IR repair stay off.

For conventional single-frame scans, NegPy selects the highest resolution reported by the device and labels it `Best quality`. TIFF and DNG contain the same scanner-linear negative data. The Process setting controls inversion, not the file format.

## Safety and recovery

The workflow includes:

- A conservative free-space check before capture.
- macOS idle-sleep prevention while a preview or batch is running.
- `Stop after current frame`, which preserves completed frames and releases the scanner safely.
- `Select remaining`, which reselects unfinished physical slots after thumbnails are loaded again.
- Slot-aware default filenames so retries do not silently renumber the roll.
- Fresh roll-fingerprint checks before fine scanning.
- Geometry, bit-depth, resolution, clipping, and stopped-transport-smear validation before import.
- Copy, flush, hash verification, and deterministic cleanup when publishing into macOS protected folders. This replaces the hard-link path that could block indefinitely.
- One terminal batch receipt recording the selected slots, completed slots, capture bundle, release result, and recovery state.

## Live hardware validation

A live roll preview returned all 40 addressable slots.

The final acceptance artifacts for slots 7, 14, and 17 contain 4000 dpi, 16-bit RGB and IR data at the expected scanner geometry. All three passed transport-smear validation and reported no clipping warning. Visual inspection showed the complete image area and rebate without a black edge or repeated-row tail.

Slots 14 and 17 completed in one live batch:

- One scanner reservation was acquired.
- Both selected slots completed.
- The scanner remained reserved between frames.
- One final release succeeded.
- No release-stall recovery was needed.
- The terminal receipt reported `complete` with no recovery required.

The same run exercised the copy-based macOS publication path without reproducing the previous hard-link stall.

## Current limits

- Hardware coverage is limited to one LS-5000 firmware revision and one converted SA-21. A stock SA-21, stock SA-30, and other Nikon models still need testing.
- The final shared-session acceptance run covered two selected frames in one batch, not an entire 36 or 37 frame roll on this exact commit.
- The live acceptance frames used directly measured positions. The inferred-position approval and fingerprint refusal paths have automated coverage but still need a dedicated live test.
- Black-and-white roll scans still open one SANE session per selected frame.
- Waking a parked feeder is best effort. Some parked or failed states still require a scanner power cycle and film reinsertion.
- The packed IR plane is real and aligned, but its firmware-side multisample behavior remains unknown.

## Verification

- `ruff check .`: passed.
- Local macOS full test suite: 2,708 passed, 4 skipped, 7 deselected, 1 expected failure, and 2 parameterized subtests passed.
- `git diff --check`: passed.
- macOS application and DMG builds: passed.
- Strict application code-signature verification: passed.
- DMG verification: passed.
- Packaged capture-bundle integrity: passed with digest `03df9baac7dcd7cb7ea60bc009609e46a47d6e18426886cced2e8cc4de9a1f8e`.
- GitHub Actions on Linux: passed for commit `a564e15f2cb8876fb76160687c28b0dea768e098` with 2,706 passed, 6 skipped, 7 deselected, and 1 expected failure ([run 29385322785](https://github.com/marcinz606/NegPy/actions/runs/29385322785)).

The implementation and recovery guide is in `docs/LS5000_ROLL_SCANNING.md`.


marcinz606:
hopefully I will be able to test it out with my upcoming coolscan during the weekend

marcinz606:
my coolscan just arrived, I will be testing this during the weekend

marcinz606:
I started testing with my LS-50 and see some issues. I'll see if i can fix them on top of your branch and make this a beginning of more generic "coolscan" route rather than ls5000

rohanpandula:
> I started testing with my LS-50 and see some issues. I'll see if i can fix them on top of your branch and make this a beginning of more generic "coolscan" route rather than ls5000

oh interesting what kind of issues are you seeing is it based on the driver or on the way negpy is using the scanner? SANE needs a lot of help from coolscan users

marcinz606:
> > I started testing with my LS-50 and see some issues. I'll see if i can fix them on top of your branch and make this a beginning of more generic "coolscan" route rather than ls5000
> 
> oh interesting what kind of issues are you seeing is it based on the driver or on the way negpy is using the scanner? SANE needs a lot of help from coolscan users

both :) one issue is 14 bit on ls-50 that I solved. Framing is off with sa-21 and few strips from different rolls that i tried. Obviously I can't try the whole roll scanning as it expects sa-30 and ls-5000 via this. On SANE end I needed to build it with a SANE_CAP_INACTIVE patch to have IR, `Archival RGB 4x + IR` is greyed out. That's what i found today

I will probably continue digging through your code and everything on coolscan during the weekend

marcinz606:
I saw your MR in sane-backends github, the SANE_CAP_INACTIVE also works on LS-50, not only 5000 :)

marcinz606:
@rohanpandula 
I had some time to look through that in more detail and tried running parts of it that I could :)

I'm not going to merge it though. Honest reason: I can't maintain it. It's almost 30k lines of device-specific code (raw libusb, captured SCSI replay plans) tuned to your exact hardware. 

I have a simple, generic  plain SANE-based coolscan improvements in the works, frame addressing, batch, eject, IR, basic 6-exposure SA-21 handling with preview. That's about the depth of scanner hardware coupling NegPy itself should carry.

What I'd love to see instead: your work as a standalone project & package on PyPI. You keep ownership and release pace (I might contribute LS-50 specific bits too in the future :) ), and if it roughly mimics the python-sane interface (open, options, scan -> array, with roll/RGBI and all other extra stuff as extensions), We can neatly plug it into NegPy as dependency same way we use python-sane or gphoto2.

Also it would benefit open source in general, not just NegPy users :) 

I'll probably have my simple, generic coolscan-sane ready and merged to main in couple of days, I'll try to do it in a way that could be easily extended with that middleware if you would go that way

rohanpandula:
@marcinz606 you know what, that's a great idea and totally fair logic. almost 30k lines of libusb replay code tuned to one firmware is a lot to ask anyone else to own, and you debugging my converted SA-21 from an LS-50 would have been miserable for both of us. I'll start cranking on the standalone package.

rough sketch of how I'm thinking about it:

core surface mimics python-sane like you said, open / options / scan -> array, so the negpy side stays a thin adapter:

```python
dev = pkg.open("ls5000")
dev.resolution = 4000
arr = dev.scan()   # 16-bit scanner-linear RGB, numpy
```

roll and RGBI live in extensions on top:

```python
roll = dev.roll()          # 40-slot addressing
roll.preview()             # 97 dpi thumbnails
f = roll.scan(7)
f.rgb, f.ir, f.receipt     # RGB + aligned IR + the scan receipt
dev.eject()
```

fingerprinting, safe stop, spacing offsets and the session/reservation logic all stay behind that. negpy never has to see a SCSI replay plan again.

transport is isolated per device, so LS-5000 fw 1.03 becomes the first entry instead of a hard assumption. would love your LS-50 bits whenever you get to them. that plus your stock SA-21 framing findings are exactly the hardware coverage I can't produce myself.

dust repair stays out of it entirely. the ICE engine is already its own package with its own test suite, so acquisition and repair end up as two separate deps negpy can adopt independently (or not). that's basically what #539 is under the hood, so the same logic applies there and I'll close it too. once the driver package exists the ICE side can come back as a thin PR that just calls into it.

one thought for your generic coolscan-sane seam: the roll workflow needs to own the scanner for a whole batch (one continuous session, one release at the end), so the main thing the extension point needs is a way to hand over the device, or just get out of the way, while a roll is in flight. whatever shape you land I'll match on my side.

closing this one out. I'd rather ship this somewhere you can actually depend on it. no package name yet, taking suggestions :)

marcinz606:
coolscanpy

PR #522, opened 2026-07-15
Reject stale render completions after fast file switches so they cannot repaint the canvas or corrupt metrics.

### LINKED ISSUE #517: Stale async render repaints the wrong frame
When you switch files quickly, a late render from the previous file can overwrite last_metrics and repaint the canvas. showing the wrong image, plus corrupt histogram / densitometer / UV-grid data, and can trigger a thumbnail write for the wrong file

jboneng:
## Summary

Fixes a race where a late async render from a previous file could repaint the canvas and corrupt metrics after fast file switches. Also makes filmstrip navigation snappy and reliable during rapid scrubbing — especially over slow network shares (NAS).

### Stale async render repaints the wrong frame

When navigating quickly, a render started for file A could finish after switching to file B. `_on_render_finished` would still emit `image_updated` and merge metrics into `last_metrics`, showing A's pixels on B's frame and corrupting histogram/bounds data.

### Navigation lag and races

Rapid filmstrip navigation had several compounding issues:

- Every hop triggered a full synchronous file commit (EXIF, DB, config hydration) on the UI thread
- Intermediate frames could start un-cancellable RAW decodes mid-burst, blocking the single preview worker
- The filmstrip red selection frame lagged behind arrow keys (paint events starved during key bursts)
- Cold landings showed a frozen canvas for 2–4+ seconds (embedded splash was extracted but only emitted after the full linear decode)
- Prefetch only warmed ±1 neighbor; active decodes shared one worker with background prefetch, so a slow read could block the frame you actually landed on

### Stale render guard

- `_on_render_finished` and `_on_metrics_updated` reject completions whose `source_hash` ≠ `current_file_hash`
- Stale renders skip repaint and metrics merge; a queued render for the current file is still dispatched
- Unit tests: `test_render_finished_skips_stale_source_hash`, `test_render_finished_stale_still_dispatches_pending_task`

### Debounced filmstrip navigation

- Split `select_file` into instant selection (`navigate_to`) + debounced heavy commit (`_activate_selection`, 300 ms)
- `selection_changed` signal for immediate filmstrip highlight updates
- `next_file` / `prev_file` / filmstrip clicks / keyboard arrows route through `navigate_to`
- Only the settled frame triggers `file_selected` → `load_file`

### Preview load coalescing

- `PreviewLoadTask.load_seq` + worker supersession: stale queued decodes are dropped without I/O
- `_preview_load_still_wanted` guards splash/finished handlers
- `load_file` clears pending render tasks and dispatches preview immediately (session-level debounce replaces controller debounce)

### Instant visual feedback

- **Early splash paint**: embedded JPEG thumbnail emitted via `on_splash` callback *before* the slow linear decode completes 
- 
- **Real-time highlight**: `sync_ui` calls `viewport().repaint()` after selection moves so the red frame tracks every arrow key during rapid nav

### Prefetch and cache (sustained scrubbing)

- **Directional prefetch**: warms next 3 frames in scrub direction + 1 behind (was ±1 only)
- **Dedicated prefetch thread**: background cache-warm decodes no longer block the active frame's decode on the main preview worker
- **Thread-safe preview cache**: `RLock` on `PreviewBufferCache` (shared by both workers)
- **Larger cache**: `preview_cache_max_entries` 8 → 24 (~480 MB of half-size previews under the 1.2 GB byte cap)
- Prefetch tasks carry `prefetch_gen` and are cancelled when navigation moves on

## Files touched

| Area | Files |
|------|-------|
| Controller | `controller.py` — stale guards, load/prefetch routing, nav direction tracking |
| Session | `session.py` — `navigate_to`, debounced commit, EXIF cache |
| Filmstrip | `files.py` — `currentChanged` driver, sync repaint |
| Workers | `render.py` — load_seq, prefetch_gen, early splash emit |
| Preview | `preview_manager.py`, `preview_cache.py` |
| Prefetch | `prefetch_logic.py` — directional window |
| Config | `config.py`, `types.py` — cache size default |

## Test plan

- [x] **Stale render**: Rapidly switch between two edited files; confirm canvas and histogram always match the selected frame (no flash of previous file)
- [x] **Rapid arrow keys**: Hold Left/Right through 10+ frames; red highlight moves on every key; canvas updates on settle (splash → full render)
- [x] **Navigate back**: Return to a recently viewed frame; should hit render memo or preview cache (near-instant)
- [x] **NAS / network share**: Load roll from `//` or slow path; scrub forward steadily — no periodic multi-second stalls from prefetch blocking active decode
- [x] **Cold landing**: Stop on a never-seen frame; embedded splash appears quickly, then sharpens to full pipeline render
- [x] **Multi-select**: Shift/Ctrl+click still builds selection without triggering single-frame navigation
- [x] `uv run pytest tests/test_controller.py tests/test_desktop_session.py tests/test_file_browser_widget.py tests/test_preview_load_coalesce.py tests/test_prefetch_logic.py -q`

fixes #517


jboneng:
After manual testing, it seems like this PR have a unintended cascading effect on multiple parts of the application.

PR #535, opened 2026-07-17
feat: consolidate crop ratio picker, add common print/screen sizes

linkmodo:
## Summary

Three related changes to the crop-ratio picker and auto-detection:

**1. Consolidated the ratio picker, added common print/screen sizes.** The dropdown listed every ratio in both orientations (3:2 *and* 2:3, 4:3 *and* 3:4, …) even though the crop tool already auto-orients to match the drag — the reversed entries never did anything a single canonical entry couldn't. Now one entry per shape, plus 7:5, 16:9, 16:10 and 8.5:11 which were missing entirely. The full `AspectRatio` enum keeps every reciprocal, because the Export "Paper ratio" picker doesn't auto-orient and genuinely needs both. `canonical_crop_ratio()` maps legacy/portrait stored values onto the visible entry.

**2. The crop box now reshapes when you pick a new ratio.** Previously it stayed visually stale until you redragged it. New `AppController.set_crop_ratio()` reshapes in place around the box's existing center, reusing `enforce_roi_aspect_ratio` (the same centered-reshape auto-crop uses). The math runs in real pixel space and accounts for 90° rotation, since a normalized "square" isn't square on a non-square image.

**3. Fixed auto-detect misclassifying 35mm scans as 7:5.** `_closest_standard_ratio` matched against the whole `AspectRatio` enum, which after change 1 included print/screen output sizes. 7:5 (1.4) and 16:10 (1.6) sit close enough to 3:2 (1.5) in log-ratio space that ordinary contour noise — sprocket holes, margin trimming — tipped a genuine 3:2 frame onto them. Detection now uses a dedicated `FILM_FORMAT_RATIOS` list of formats a camera/scanner can actually produce; the picker keeps the full list.

**4. Fixed a colour shift on ratio change.** Changing the ratio visibly shifted colour on the canvas. `set_crop_ratio` was clearing the per-file metering bounds, and since normalization meters *within* the crop ROI (`resolve_analysis_region`), re-analyzing over the new smaller ROI landed on different per-channel floors/ceils. On a real scan going 3:2 → 4:3 the red floor moved `+0.0058` while blue moved `-0.0056` — opposite directions, a cast shift rather than a brightness change. The other crop paths clear the bounds because the crop decides whether the film rebate is inside the metered region; a ratio change can't pull rebate in, since both the reshape and autocrop's `_enforce_ratio_by_occupancy` only ever shrink within a footprint that already excludes it. So the invalidation there could only ever drift. **This predates the branch** — with auto crop on, the ratio has always fed `get_autocrop_coords` and re-metered; change 2 extended the same fault to the manual-crop path.

## Test plan
- [x] **Colour shift**, verified against the real app on both the manual-crop and auto-crop paths: rendered floors/ceils are now bit-identical across a ratio change, and the shift reproduces when the invalidation is put back
- [x] **Reshape**, verified against the real app: a 0.6×0.7 box at (0.4, 0.55) becomes a true 747×747 pixel square on a 1600×1067 preview, center preserved
- [x] **Detection**: regression test sweeps ROIs measuring 1.42–1.58 (a real 3:2 frame plus plausible detection noise) and confirms none flip to 7:5/16:10
- [x] **Picker**, verified against the real app: dropdown matches `CROP_RATIO_CHOICES` with no reciprocal duplicates, and a legacy `"2:3"` stored value displays as `"3:2"`
- [x] 8 new unit tests across `test_controller.py` and `test_geometry_logic.py`, including a guard that the reshape never grows the box (the invariant change 4 depends on)
- [x] Full suite: 1963 passed. The one failure, `test_export_presets.py::test_output_dir_subfolder_of_source`, is a pre-existing Windows path-separator issue that reproduces on a clean `main`
- [x] `ruff check` / `ruff format --check` clean

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -41,6 +41,7 @@ from negpy.domain.models import (
     ExportPresetOutputMode,
     ExportResolutionMode,
     WorkspaceConfig,
+    canonical_crop_ratio,
     export_blocked,
     flat_export_config,
     flat_master_config,
@@ -1309,7 +1310,11 @@ class AppController(QObject):
         if geom.fine_rotation != 0.0:
             transformed = apply_fine_rotation(transformed, geom.fine_rotation)
 
-        new_ratio = detect_closest_aspect_ratio(transformed, fallback=geom.autocrop_ratio)
+        # Detection can match a portrait-oriented frame to a portrait-only AspectRatio
+        # (e.g. "2:3") that the ratio picker doesn't display — canonicalize so the
+        # stored ratio always matches an entry the picker can show (see
+        # domain.models.CROP_RATIO_CHOICES; the crop tool auto-orients regardless).
+        new_ratio = canonical_crop_ratio(detect_closest_aspect_ratio(transformed, fallback=geom.autocrop_ratio))
         if new_ratio == geom.autocrop_ratio:
             return
 
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -55,7 +55,7 @@ from negpy.features.exposure.logic import (
 )
 from negpy.features.exposure.models import ExposureConfig
 from negpy.features.finish.models import FinishConfig
-from negpy.features.geometry.logic import apply_fine_rotation, detect_closest_aspect_ratio
+from negpy.features.geometry.logic import apply_fine_rotation, detect_closest_aspect_ratio, enforce_roi_aspect_ratio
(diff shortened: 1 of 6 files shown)


marcinz606:
nice, I wanted to consolidate it for a while

linkmodo:
Added enhancements on top of crop ratio consolidation to auto change crop ratio on canvas, and fixed a bug where cropping an image induced color edits.

PR #539, opened 2026-07-17
Scan roll frames for Digital ICE and repair them with the portable engine

rohanpandula:
## Summary

This PR connects #497's roll scanning to [Portable Digital ICE](https://github.com/rohanpandula/digital-fauxice), an independent infrared dust and scratch repair for the Nikon LS-5000 that I reconstructed from the scanner's own behaviour. Its output is byte-exact against the original implementation on two complete validation frames. Frames selected on the roll contact sheet get captured for the engine and come back as cleaned masters, with an optional inpainting stage for the regions too damaged for exact repair.

Three layers, bottom up.

The acquisition boundary captures the exact two-source input the engine requires: a 285 dpi single-sample RGBI prepass, then a locked 4000 dpi single-sample RGBI main scan through one SANE handle, with the prepass focus and exposure replayed and re-verified. It writes an immutable, self-verifying capture bundle and refuses anything other than an LS-5000 before changing scanner state.

The processing adapter is lazy and optional. It exposes `off`, `cpu`, `cpu-fast`, `metal`, and `cuda`, and deliberately leaves out the engine's automatic mode. A backend that cannot start fails with its reason instead of quietly becoming a different backend. The engine receipt is verified (output hash, profile ID, pixel counters, RNG state) before the cleaned RGB is trusted.

The roll workflow is where the two meet. The #497 contact sheet gains a "Digital ICE (dual RGBI)" toggle and backend selector for color negative material. Each selected slot becomes one complete cycle: the reviewed roll fingerprint is re-verified, the slot's Film Spacing Offset resolves through the existing coolscan3 geometry translation (with a new guard that refuses an offset that crossed a slot boundary rather than scanning the wrong frame), one dedicated SANE session captures the pair and closes. Only after the scanner is released does the engine repair the frame. NegPy then publishes one cleaned 16-bit scanner-linear TIFF beside a `_SCAN.json` receipt recording the slot, offset, plan, bundle hash, backend selection, and the engine's full receipt. Cleaned frames import as C-41 with the file-based IR heal left off, since their dust is already repaired.

A processing failure preserves the frame's verified bundle on disk and names it in the error, so the repair can be retried without a rescan. Completed frames stay published and "Select remaining" keeps working. Stop takes effect between frames, not in the middle of a hardware transaction.

## Hybrid repair

Some defects are too broad for exact repair: deep gouges, large debris, damage crossing strong edges. An optional "Hybrid repair" toggle routes exactly those regions to a bounded LaMa inpainting fallback while the exact result stays in place everywhere else. Infrared detection remains in charge, the routed area is capped at 2% of the frame, and outside the synthesis mask the output is byte-identical to the pure repair by the tool's own verified contract.

Each hybrid frame publishes four files with one base name: the hybrid master (`.tif`), the pure ICE master (`_ICE.tif`, preserving the non-generative reconstruction inside the mask), the synthesis disclosure mask (`_SYNTH.png`), and a receipt binding all three by hash with the hybrid tool's own receipt embedded. The hybrid tool ships in the engine repo under `hybrid/`; NegPy drives it as a subprocess configured entirely through the environment, and the checkbox stays disabled with a reason when that runtime is not set up. Both validation frames were reproduced byte-for-byte through this exact path before it landed here.

## Backends

| Backend | Time per 4000 dpi frame | Host |
| --- | --- | --- |
| `cpu` (reference) | roughly an hour | Apple M4 |
| `cpu-fast` | 9.2 to 9.5 s | Apple M4, default thread count |
| `metal` | about 8 s | Apple M4 GPU |
| `cuda` | 5.3 to 5.8 s | NVIDIA RTX A4000 |

`cpu-fast` and `metal` are byte-identical to the reference, proven at startup on a synthetic job; if that proof fails the backend refuses to run. The worker runs the same self-test before opening the scanner, so a session is never spent on a backend that would have refused to start. The Metal backend executes every double-precision operation through a software IEEE-754 implementation in integer arithmetic, since Apple GPUs have no float64 hardware; its complete-frame parity receipts are in the engine repo. The `fast` extra accepts numba 0.65, the version NegPy already pins.

## Install

The engine is optional. Nothing here runs without it, and the UI says so with an install pointer.

```sh
python -m pip install \
  "portable-digital-ice[metal] @ https://github.com/rohanpandula/digital-fauxice/releases/download/v0.3.0/portable_digital_ice-0.3.0-py3-none-any.whl"
```

Details, including the CUDA and compiled CPU extras and the hybrid runtime setup, are in `docs/PORTABLE_DIGITAL_ICE.md`.

## The trade

The engine's exactness claim is anchored to what the scanner's original software captured: single-sample scans. #497's RGB 4x averaging changes the noise statistics the engine's reverse-engineered constants were calibrated against, so an ICE batch gives up the 4x averaging. The panel states this when the toggle is on. When the archival multi-sample master matters more than exact ICE repair, run a normal RGB4x+IR batch. Nothing about #497's own path changed.

## Current limits

- The engine's byte-exact complete-frame receipts cover two independent mounted C-41 frames. Roll-position acquisition is implemented and gated by the same evidence machinery, but roll geometry has not yet passed its own frozen full-frame parity gate. Until it does, roll ICE output should not carry the exactness label, and the docs say so.
- The roll ICE cycle has full automated coverage with injected fakes but has not yet run against live hardware end to end.
- Inpainting runs on the CPU. The pinned IOPaint 1.6.0 blocklists LaMa on Apple's MPS and would silently run on CPU while reporting otherwise, so the adapter refuses that device rather than misreport it.
- Infrared-compatible color film only. Silver black-and-white film and some Kodachrome block infrared in the image itself, which rules out any infrared cleaner, this one included.
- The engine has had no third-party security audit.

Nikon, COOLSCAN, and Digital ICE are trademarks of their respective owners, used here only to identify the hardware and behaviour the engine was studied against. The engine is an independent implementation, not affiliated with or endorsed by Nikon. No Nikon executable, library, firmware image, or proprietary profile is redistributed.

## Validation

- Full NegPy suite on this branch: 2,782 passed, 5 skipped, 7 deselected, 1 xfailed, 0 failed.
- Digital ICE suites with the real engine importable: pass, including contract tests asserting byte-identical adapter output across backends and the hybrid orchestration seams.
- The hybrid tool's own suite: 312 passed, and both validation frames reproduced byte-for-byte through the published package before it landed here.
- `ruff check .` and the CI `ty` command: passed.
- Engine source, the v0.3.0 release, and the parity receipts for every backend: [rohanpandula/digital-fauxice](https://github.com/rohanpandula/digital-fauxice).

This branch is based on #497. Until that PR merges, GitHub shows its changes in this diff as well.


rohanpandula:
closing in favor of the standalone package route, discussion in #497. the ICE engine comes back as a dependency plus a much smaller integration PR once the driver package is up

PR #546, opened 2026-07-18
fix: persist dodge/burn mask visibility across image switches and restarts

### LINKED ISSUE #540: Mask visibility setting does not persist between launches or image switches.
When dodge/burn masks are set up, they always default to visible when switching between images in the session and on restart of the application.

A better UX would be to have that setting persist at whatever the user set it to last for that image.

New masks should always default to visible, of course.

paulglover:
## Summary

Fixes #540 — dodge/burn mask visibility now persists across image switches **and** app restarts.

Previously the per-mask "hide outline" toggle lived in a session-only `local_hidden_masks` set keyed by mask index into the *current* file, and it was cleared on every file switch. So visibility reset to "all shown" whenever you changed images, and never survived a restart.

## What changed

- **Per-file, persisted store.** Hidden state now lives in `AppState.local_hidden_masks_by_hash` (content hash → set of indices), written through to the `hidden_masks_by_hash` global setting on every toggle and reloaded on launch. It is deliberately kept out of `WorkspaceConfig`, so toggling an eye icon never dirties the edit/undo history or writes a `.negpy` sidecar — it's view-only state and never affects render or export.
- **Single source of truth.** `local_hidden_masks` is now a derived property over that store (getter returns the current file's set; setter writes it). Switching files needs no special restore step, and the manual two-copy sync that caused the original bug class is gone.
- **Stale-index clamp.** The getter returns a fresh copy clamped to the current mask count, so a config swap that shrinks the mask list (undo/redo/jump-to-step) can't surface an out-of-range index. Storage self-heals on the next toggle.

## Testing

- `make all` green — 1990 passed, 4 skipped.
- New tests: per-file persistence, restart round-trip, empty-set pruning, and the stale-index clamp.

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -1553,10 +1553,13 @@ class AppController(QObject):
         """Show/hide one mask's outline on the canvas (view-only; no re-render)."""
         if not (0 <= index < len(self.state.config.local.masks)):
             return
+        hidden = set(self.state.local_hidden_masks)
         if visible:
-            self.state.local_hidden_masks.discard(index)
+            hidden.discard(index)
         else:
-            self.state.local_hidden_masks.add(index)
+            hidden.add(index)
+        self.state.local_hidden_masks = hidden
+        self.session.persist_hidden_masks()
         if self.canvas:
             self.canvas.overlay.update()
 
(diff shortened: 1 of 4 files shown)


marcinz606:
merging, thanks!

paulglover:
No problem! This is shaping up to be a really useful tool for film-digital hybrid work. Glad to be able to help in some way!

PR #554, opened 2026-07-18
feat(assets): half-frame mode — two frames per scan

### LINKED ISSUE #553: half-frame mode
mode for half frame scans when you scan the diptych but can process both separately

marcinz606:
## What

Half-frame cameras (Pentax 17, Olympus Pen…) expose two photos per 35mm frame; scans arrive as one file with two images separated by an unexposed gutter. This adds a **Half Frame** toggle (Files toolbar, mirrors RGB-scan end to end): each scan appears as two frames on the contact sheet, each with independent edits, bounds metering, thumbnails, sidecars, and exports.

## How

- **Identity**: discovery expands each file into two assets with hash `<base>#1` / `#2` (`negpy/services/assets/half_frame.py`). Every hash-keyed store — `file_settings`, `edit_history`, `file_marks`, disk thumbnails, EXIF, render memo — becomes per-half with no schema changes.
- **Independent bounds**: the decoded buffer is sliced at the gutter before the pipeline, so full-frame metering *is* per-half metering. No shader changes; GPU caches keyed by the suffixed `source_hash`.
- **Shared decode**: decode/preview caches key on the unsuffixed hash (`base_hash()`), so both halves reuse one decode. Slices are copied — cached buffers stay read-only.
- **Gutter auto-detect**: 1-D column-luma heuristic with three gates — local-median deviation (immune to frame-brightness differences), extremal-vs-both-sides (rejects in-scene step edges), full-height uniformity (rejects textured features). Center-split fallback; the crop tool trims residue.
- **Collision guards**: per-half sidecars `name.N.negpy`; export stems `name_1`/`name_2`; `add_files` same-path guard matches per half; EXIF path-fallback rehoming excludes `#` hashes so it can't steal a half's edit.
- Toggling off re-collapses to whole scans; per-half edits stay in the DB and return on re-enable.

## Validation

- 19 new tests (`tests/test_half_frame.py`): detector polarities/fallbacks/rejections, slice partition, identities, sidecar/export suffixes, discovery expansion, dedup guard, per-half bounds independence.
- Detector validated on all 7 `samples/half_frame` Pentax 17 scans: 6 split dead-center in the gutter; IMG428 (invisible gutter, night frames) correctly falls back to center.
- Headless app run (Xvfb): 7 files → 14 filmstrip entries, both halves of one scan rendered independently at half-width with distinct stats.
- `make all` green — 2022 passed.

## Out of scope (v1)

Divider-drag UI, per-file split (mode is global like RGB-scan), RGB-triplet + half-frame combination (triplets stay whole).

closes #553

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,6 +64,10 @@ Every feature lives in `negpy/features/<name>/`:
 5. Add a sidebar and register it in `ControlsPanel`
 6. Add unit tests; if the feature has both CPU and GPU paths, add a parity test (pattern: `test_gpu_curve_parity.py`)
 
+## Style
+
+- **Comments minimal.** Comment only non-obvious constraints the code can't express (a cache contract, an ordering requirement, a rejected-alternative trap). No comments that narrate what the next line does, restate the diff, or justify a change to a reviewer. Prefer one dense line over a paragraph; docstrings short and factual.
+
 ## Invariants & gotchas
 
 - **CPU/GPU parity**: any change to a stage's math must land in both `logic.py` and its `.wgsl` shader. Constants mirrored as WGSL literals (histogram bins, zone density, metrics offsets) have parity tests — keep them in sync.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 0.39.0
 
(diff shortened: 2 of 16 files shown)


ycjhjiangxiao-commits:
Can you extend this function to split film strip scan into multiple frames?

PR #556, opened 2026-07-19
feat(scan): Coolscan / SANE film scanning

marcinz606:
## Summary

Adds direct **SANE film-scanner support** to the Scan tab, built around the Nikon Coolscan family (LS-50 / Coolscan V, LS-5000) and best-effort for any SANE film scanner. Verified live on a Nikon LS-50 + SA-21 strip adapter.

You can select/refresh/eject a device, pick DPI and bit depth, run a frame-range batch, and set the scan window visually through a whole-strip preview dialog. Behind it, the scanner layer was reworked into a backend-neutral seam so device-specific drivers can live out of tree.

## What's in it

**Scanning**
- Device select / refresh / eject, DPI + bit depth (14-bit-in-16-bit container fix), autofocus, hardware auto-exposure.
- Frame-range batch with a display-sleep power assertion so long strips don't stall.
- **Strip preview dialog**: previews the whole strip at low DPI, negative→positive, one draggable crop window reused for every frame, per-frame selection (skip frames), a feed-axis offset with a live indicator, grid layout, and a drift slider (mm/frame) to track cumulative feed error across a strip. The window is stored as a normalized fraction and mapped to the backend's native units.
- Eject shells out to `scanimage --eject` (python-sane can't press a SANE button); USB re-enumeration self-heal; transient-I/O retry.

**Scanner seam**
- `ScannerBackend` / `ScannerSession` protocol states the real contract; backends classify their own failures (`TransientScanError`) instead of the service matching SANE status strings.
- `RollSession` / `PerFrameRollSession` for whole-strip preview as an iterator; offsets cross the seam as a fraction of one frame pitch.
- `tests/scanners/test_backend_contract.py` is a compliance suite an out-of-tree backend can parametrise against itself.

**IR dust removal** — fixed to work on real scanner IR planes: the holder-mask margin no longer reads as one frame-sized defect, and the B&W/Kodachrome bail now keys on the fitted crosstalk exponent instead of raw ratio dip. Verified on a six-frame roll (203–512 defects/frame, gain median 1.0008).

## ⚠️ Requires patched SANE (IR + correct SA-21 spacing)

Two things do **not** work against stock `sane-backends` and need a patched `coolscan3` build:

1. **Infrared dust channel** — stock `coolscan3` compiles IR out (`#if 0 SANE_FRAME_RGBI`); it must be un-gated.
2. **Correct frame spacing on the SA-21 strip adapter** — stock uses the LS-30-era 1.5in frame advance (38.106 mm). The true 135 pitch is 38.0 mm (8 × 4.75 mm perforations); the wrong value drifts ~−0.28 mm/frame and eventually collapses offset scans to empty. The fork corrects it (A/B-measured −0.277 → −0.143 mm/frame on an LS-50).

Until proper fixes land in mainline libsane use rohanpandula's `coolscan3-ls5000-38mm-frame-pitch` branch. `make sane-rgbi-help` prints build steps; `make run-ir` launches the app against the patched build (nothing system-wide is touched). Note: the pitch fix is gated `CS3_TYPE_LS5000` only — **on an LS-50 you must widen the type gate locally** (documented in `sane-rgbi-help`). The drift slider is the runtime fallback when running against unpatched SANE.

## Tests

New coverage across `tests/scanners/` (backend contract, capabilities, eject, SANE session/reopen/window/depth, IR) and desktop-side (`test_scan_worker`, `test_scan_sidebar`, `test_strip_preview_dialog`, `test_power_assertion`, `test_ir_dust`). `make all` green.

## Status

Fine per-frame windowed scanning stays on `run_batch`; the whole-roll seam is settled for preview but the native windowed transport is still owed. Hardware passes done on LS-50 + SA-21.


marcinz606:
@rohanpandula draft/wip of basic sane implementation I got to so far, There is some tweaking to do left especially when it comes to offset settings when scanning the strip

rohanpandula:
funny timing :) a lot of this shipped on my side this week.

on offsets: I think your drift slider is compensating coolscan3 itself. the backend's frame advance is still an LS-30-era guess (`resy_max * 1.5 + 1`, the comment literally says "maybe not for others"), while a 135 frame advances 8 x 4.75mm = 38.0mm, which is what you measured (37.96-38.00). just pushed two branches to my sane fork: `coolscan3-ls5000-38mm-frame-pitch` (the constant, with the perforation math) and `single-pass-rgbi4` (on top of the RGBI MR you saw: reads the scanner-declared boundary instead of any constant, plus packed single-pass RGBI). the load/eject zeroing already merged upstream. with the pitch right, per-frame drift should mostly vanish.

coolscanpy 0.1.3 also grew short-strip support: per-frame spacing from the preview fingerprint, RefeedRequired when a strip parks at the end-stop. the live roll preview came back aligned with no manual spacing offset and correctly flagged the trailing sliver.

session seam is exactly what I meant in #497, I'll build the adapter against it once this merges. happy to run this branch on the 5000 too.

marcinz606:
yup, i took a bunch from your MR from last week :)

marcinz606:
@rohanpandula
I've spent some time trying to make our scanner layer less SANE-shaped so coolscanpy has something sane :D to plug into.

main things: ScannerBackend declares what the service calls now, open_session hands back a protocol instead of the concrete SaneSession, and backends decide for themselves what's retryable, before, the service was grepping SANE error strings.
Also added a RollSession for whole-strip preview. it's an iterator, so we can yield per frame and you can do your one traversal and yield from memory.

does that look workable from your end, or would you do it differently?

PR #579, opened 2026-07-21
feat: DB management dialog to view and clear stored data

linkmodo:
## Summary

There was no in-app way to inspect or wipe the SQLite databases, so a reloaded image always restored its previously saved edits with no way to opt out short of hand-deleting the `.db` files.

Adds a **"Manage Database…"** entry to the More Actions (⋯) menu that opens a dialog:

- **View** everything stored, grouped as *per-image* (saved edits, undo history, keep/reject marks) and *tooling* (normalization rolls, flat-field profiles, export presets, app preferences), plus total on-disk size including the WAL/SHM sidecars.
- **Clear Saved Edits** — drops the per-image looks + history + marks so a reloaded image starts from defaults; export presets and rig profiles survive.
- **Reset Everything** (destructive/red) — wipes both databases to a first-run state.

Both clears sit behind a warning dialog that defaults to Cancel and says "This cannot be undone". Counts refresh in place afterward and buttons disable when there's nothing left to clear.

## Implementation notes

- Repository gains `database_stats()`, `clear_saved_edits()`, `reset_everything()`.
- Clears run `PRAGMA wal_checkpoint(TRUNCATE)` + `VACUUM` so the on-disk footprint actually shrinks, and delete **rows only** (schema preserved) so the app keeps working against the emptied databases without a re-init or restart.
- `export_presets` is one JSON row inside `global_settings`, so it's counted from that list and excluded from the "app preferences" count to avoid double-counting.
- Clearing is **DB-only** — it never touches source files or the thumbnail cache. The dialog notes that if `.negpy` sidecars are enabled, those still sit next to source images and can restore an edit on reload (sidecars are off by default).

## Test plan
- [x] Verified end-to-end against the real desktop app: menu entry present; seeded 142 edits / 40 history steps / 38 marks / presets / profiles / prefs and confirmed the dialog counts; ran both clears through the real dialog and confirmed the correct categories were dropped/kept; repo still writable after a full reset (screenshots captured)
- [x] 7 new unit tests: per-category stats, the export-presets-vs-preferences split, targeted vs full clear, the reloaded-image-starts-fresh goal (`load_file_settings` returns `None` after clear, path fallback doesn't resurrect it), post-reset usability, and empty-DB no-op — each confirmed to fail against a sabotaged no-op clear
- [x] Full suite: 2352 passed. The 3 failures are pre-existing, unrelated Windows path-separator issues that reproduce on a clean `main`
- [x] `ruff check` / `ruff format --check` clean

### DIFF
--- a/negpy/desktop/view/canvas/toolbar.py
+++ b/negpy/desktop/view/canvas/toolbar.py
@@ -262,6 +262,9 @@ class ActionToolbar(QWidget):
         )
         overflow_menu.addSeparator()
 
+        overflow_menu.addAction(qta.icon("fa5s.database", color=icon_color), "Manage Database…", self._show_database_dialog)
+        overflow_menu.addSeparator()
+
         overflow_menu.addAction(qta.icon("fa5s.map-signs", color=icon_color), "Take the tour", self._show_tour)
         overflow_menu.addAction(qta.icon("fa5s.keyboard", color=icon_color), "Keyboard Shortcuts  ?", self._show_shortcuts)
         self.btn_overflow.setMenu(overflow_menu)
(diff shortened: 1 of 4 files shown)


marcinz606:
really good idea, merging

PR #583, opened 2026-07-21
SANE/pieusb: RGBI + NegPys IR dust removal

hullrich:
While looking at the coolscan integration, i realized `_reinterpret_channels()` can also be used for pie-usb scanners to extract the 4th channel in RGBI mode.
With this pieusb can now also use NegPys IR Dust removal (which is miles better than the internal one).

I am aware that NegPy is moving towards a SANE-less integration, but this PR would streamline the pieusb scanning in the mean time.

### DIFF
--- a/negpy/infrastructure/scanners/sane_backend.py
+++ b/negpy/infrastructure/scanners/sane_backend.py
@@ -856,7 +856,7 @@ class SaneBackend:
             # Not every backend has a `mode` option (coolscan3 exposes none) —
             # only touch options the device has.
             if hasattr(dev, "opt") and "mode" in dev.opt:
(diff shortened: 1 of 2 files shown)


hullrich:
For reference, this is a continuation of #293

marcinz606:
I'm not moving away from sane, just opening towards different options :) I obviously have no way of testing it, I assume you verified that IR channel is properly extracted and NegPy IR removal works with it?

hullrich:
Yes i tested the 4th channel extraction in a separate script (IR Channel looked fine) and then adjusted the IR-Strategy for pieusb accordingly.
The removal by NegPy is waaaaay cleaner than the previous solution. :D

Issue #617, opened 2026-07-24
Live view fails on Fujifilm GFX50S II — silent I/O timeout + UI hang, plus a separate SIGSEGV crash in CameraWidget.get_value


noisy-ctrl:
I continue to love you work. Updates are great! I get a little excited with your momentum. 

Similar to before, #478 , app crashes on OSX when attempting to live view with my GFX 50SII. 
Apologies for the late reply (before issue ticket closed). Will watch for replies, submit any logs needed if it is helpful. 

Courtesy of Claude (I don't understand enough to be a contributor, but happy to test and find bugs): 

**Live view fails on Fujifilm GFX50S II — silent I/O timeout + UI hang, plus a separate SIGSEGV crash in CameraWidget.get_value
Summary**

Live view consistently fails to start on a Fujifilm GFX50S II. The backend correctly detects the failure and drops the session, but the UI never reflects this and is left permanently stuck on "loading live view." Investigation traced a confirmed duplicate libusb load, caused by NegPy's bundled SANE library loading scanner backend plugins from a Homebrew installation instead of anywhere inside the app bundle — resulting in two independent libusb contexts running in the same process. This was tested and ruled out as the cause of the live view failure: with the duplicate load fully eliminated (Homebrew sane-backends removed, confirmed via lsof), live view still fails with the identical -110 I/O in progress timeout. The duplicate-library issue is still worth fixing as a real structural problem, but it is a separate bug from the live view failure, not its cause. Separately, a hard crash (SIGSEGV) has also been observed inside python-gphoto2's CameraWidget.get_value() during a capture/live-view session with the same camera; whether that's related to the live view timeout is unconfirmed.

Environment
NegPy version: 0.0.0 (as reported in crash log — version string appears unset in this build)
OS: macOS 26.3.1 (build 25D2128)
Hardware: Mac16,11 (Apple M4 Pro, Apple Silicon / ARM-64)
Camera: Fujifilm GFX50S II
Connection: USB-C, direct to Mac (no hub) — also reproduced through multiple different cables and ports
GPU backend: WebGPU via Metal (unrelated to this issue, included for completeness)
Bug 1 — Live view times out and UI hangs indefinitely

Reproduction steps:

Connect GFX50S II via USB.
Open NegPy, start live view.
UI shows "loading live view" and never resolves — no error, no timeout message, no way to recover short of restarting the app.

What's actually happening (from negpy.log, timestamps consecutive, ~1 second apart):

[...]

PR #630, opened 2026-07-25
Camera scanning: check the preview capability before starting live view (fixes #621)

### LINKED ISSUE #621: Live view on bodies without CAPTURE_PREVIEW wedges the PTP session — blocks all scanning (Sony a6000)
## Summary

NegPy starts live view without checking whether libgphoto2 says the body supports it. On a
Sony a6000 — where libgphoto2 explicitly reports no preview capability — `capture_preview()`
fails **and leaves the camera's PTP session wedged**, requiring a physical power cycle before
the body will connect again.

Because the only Scan entry point opens the Live View window, this makes the camera unusable
for scanning entirely, not merely missing a preview pane.

libgphoto2 is behaving correctly here and publishes the capability up front. NegPy just never
reads it.

## Environment

| | |
|---|---|
| NegPy | 0.41.0 (`07dd965`) |
| OS | macOS 27.0 (26A5378n), Apple Silicon / arm64 |
| Python | 3.13.5 |
| python-gphoto2 | 2.6.4 |
| libgphoto2 | 2.5.34 |
| Camera | Sony a6000 (ILCE-6000), USB `0x054c:0x094e`, PC Remote mode, direct USB |

## What libgphoto2 declares

```python
import gphoto2 as gp
al = gp.CameraAbilitiesList(); al.load()
ab = al.get_abilities(al.lookup_model("Sony Alpha-A6000 (Control)"))
print(ab.operations)  # 49
```

`49` = `CAPTURE_IMAGE(1) | CONFIG(16) | TRIGGER_CAPTURE(32)`.
`GP_OPERATION_CAPTURE_PREVIEW` (8) is **absent**.

This is not an upstream oversight. Sweeping the full ability list, the a6000 is the only body
in its line without the preview bit:

| Body | Preview |
|---|---|
| a6000 | ❌ |
| a6100 / a6300 / a6500 / a6600 / a6700 | ✅ |
| a7 / a7r (gen 1) | ❌ |
| a7 III / a7 IV / a7S / a7C II | ✅ |

## Reproduction

1. Connect an a6000 in PC Remote mode. Confirm it connects (`Camera().init()` succeeds and
   `iso` / `shutterspeed` / `f-number` read real values).
2. Open **Live View & Scan**.
3. Live view never appears. The camera is now unresponsive.
4. Every subsequent connection attempt fails with `-10` until the body is physically
   power-cycled.

Minimal reproduction without the UI:

```python
import gphoto2 as gp
c = gp.Camera(); c.init()
print(c.get_single_config("iso").get_value())   # '100' — session healthy

try:
    c.capture_preview()
except gp.GPhoto2Error as e:
    print(e.code, e)                            # -7 [-7] I/O problem

c.exit()

# every attempt from here returns -10 until the camera is power-cycled
for i in range(4):
    try:
        gp.Camera().init(); print(i, "OK")
    except gp.GPhoto2Error as e:
        print(i, e.code, e)                     # -10, -10, -10, -10
```

Note the failure is `-7` (I/O problem), not `-6` (NOT_SUPPORTED) — libgphoto2 forwards the
call and the body refuses it at the device level. That refusal is what wedges the session.

## Root cause

`GphotoCamera.open()` / `start()` (`negpy/infrastructure/capture/gphoto.py`) never inspect
`abilities.operations`. `start()` spawns the preview thread unconditionally, and
`_preview_loop()` calls `capture_preview()` on a body that has already advertised it cannot
do it.

`_MAX_PREVIEW_FAILURES` then retries **three times** against a body that is already wedged by
the first call, compounding the damage before the session is dropped.

## Suggested fixes

1. **Check the capability before starting** — in `start()`, read
   `abilities.operations & GP_OPERATION_CAPTURE_PREVIEW` and, if absent, don't spawn the
   preview thread at all. Surface "this body doesn't support live view" instead. This
   *prevents* the wedge rather than reporting it afterward.
2. **Don't retry a wedged body** — once `capture_preview()` fails at the device level,
   further attempts cannot succeed and make recovery worse.
3. **`docs/CAMERA_SCANNING.md:116`** currently attributes a black live view to the body
   dropping out of PC Remote and advises a power cycle. That is the correct *remedy* here but
   the wrong *diagnosis* — it implies the camera will work after cycling, which it won't.
4. **Consider decoupling capture from live view.** `CAPTURE_IMAGE` and `CONFIG` are both
   supported on this body, but the only Scan button lives inside the Live View window
   (`live_view_window.py:140`, reached via `lv_btn` in `scanlight.py:223`), so there is no way
   to use a body that can capture but not preview. A non-live-view capture path would make
   these bodies usable instead of blocked.

## Relationship to #617

Related but a distinct root cause, so filing separately rather than adding to that thread.

| | #617 (Fujifilm GFX50S II) | This issue (Sony a6000) |
|---|---|---|
| Body supports live view? | Yes — works in other tethering apps | No — libgphoto2 reports no preview bit |
| Failure | `-110 I/O in progress` at runtime | `-7` refusal, then session wedged |
| Fix | Propagate the runtime failure to the UI | Don't attempt it in the first place |

Same visible symptom, opposite ends of the pipeline. #617 asks for a failure to be surfaced;
this asks for it not to happen. Fixing #617 alone would still leave the a6000 wedging on every
attempt.

## Not verified

- Whether other non-preview bodies (first-gen a7/a7r, a6000 in MTP mode) wedge the same way —
  only the a6000 in PC Remote was tested here.
- Whether the wedge originates in libgphoto2's Sony path or in the camera firmware. Only the
  observable behaviour is reported above.
- One incidental observation, not investigated: Sony properties return placeholders for a
  short window after `init()` (`iso='0'`, `f-number='f/0'`, `shutterspeed='Bulb'`,
  `batterylevel='253%'`, several `Unknown value 0000`) before settling to real values. Reading
  camera settings immediately after connecting may pick these up. Possibly worth a separate
  look; unrelated to the bug above.

---

*Investigated with Claude Code on the hardware described above. Every command and its output
in this report was actually run against the camera — but I'd rather you re-run the
reproduction than take my word for it. Happy to test patches or gather more logs; the a6000
is on hand.*

light-sntchr:
Fixes #621.

On a body whose libgphoto2 entry has no `CAPTURE_PREVIEW`, NegPy started the preview thread anyway. The Sony a6000 refuses `capture_preview()` at the device level, and that refusal wedges its PTP session until the camera is power-cycled, so the retry loop made it worse three times over. Because the Scan button lives inside the live-view window, the body was then unusable for scanning at all, not merely missing a preview.

## What changed

- `GphotoCamera` reads `Camera.get_abilities()` once per session. `start()` raises a typed `LiveViewUnsupported` before the first `capture_preview()` call, so the wedge is prevented rather than reported afterwards.
- The message distinguishes two cases, which matters more than it first looks. Most bodies appear twice in the database, once per USB mode, and the MTP variant drops the preview bit while the Control/PC-Remote one keeps it (a7 IV, a6300 and others). A missing bit under MTP is therefore a wrong USB mode the user can fix, not a limit of the camera. Only a body that lacks it in its tethering mode, like the a6000, genuinely cannot preview.
- The scan window still opens and still scans; only the preview pane is replaced by the reason. Hiding the window would remove the single Scan button and lock these bodies out, which is the situation the report describes.
- Calibration is greyed out with a reason, since it aims the base ROI through the live frame.
- If a body advertises no `CONFIG` either, the RGB triplet path is switched off and only plain white-light scanning is offered, because shutter, ISO and aperture cannot be written.
- `docs/CAMERA_SCANNING.md` gains the case. The existing "live view is black" row stays, but it no longer has to carry a diagnosis it never fitted.

## Deliberate limits

- **A missing bit is evidence, a present one is not.** Unlisted bodies match the generic `USB PTP Class Camera` entry, which advertises every operation. An a7C II matches it and works perfectly, so gating on "not in the database" would disable a working setup. Only absence is acted on, and a failed abilities read is treated as full support.
- **The three preview retries stay.** The report suggests dropping them, but they are what distinguishes one dropped frame from a dead session in the GFX50S II case in #617, and the a6000 no longer reaches that loop at all.
- The settings JSON is normally published by the preview loop, so a body that never streams would leave the exposure steppers empty and the preset's labels unresolvable. It is now published once when live view is refused, and the scan window applies the preset exposure at that point.

Verified against the abilities database shipped with python-gphoto2 2.6.4 (libgphoto2 2.5.34): a6000 reports 49, so `CAPTURE_IMAGE | CONFIG | TRIGGER_CAPTURE` without preview, while the a6300 reports 57 with it, matching the report exactly. Unit tests cover the refusal, the MTP wording, the generic-entry case and an unreadable abilities read. Regression-tested on an a7C II, which matches the generic entry: live view, calibration and RGB scanning are unaffected. `make all` is green with and without the optional camera group.

I cannot test an a6000 myself, so confirmation from @narrowstacks that the session no longer wedges would be welcome.


narrowstacks:
Confirming this fixes #621 on the hardware that reported it. Tested the `camera-capability-gating` branch against the a6000 directly.

**Environment:** Sony a6000 (ILCE-6000), PC Remote mode, direct USB, macOS 27.0 / Apple Silicon, python-gphoto2 2.6.4, libgphoto2 2.5.34.

**Capability read matches the report.** The body enumerates as `Sony Alpha-A6000 (Control)` with `operations = 49` — `CAPTURE_IMAGE | CONFIG | TRIGGER_CAPTURE`, no `CAPTURE_PREVIEW`. The new `_read_capabilities` picks that up correctly:

```
gphoto2 session open: ILCE-6000 (driver 'Sony Alpha-A6000 (Control)', live view NO, settings yes)
```

**The session no longer wedges.** `start()` raises `LiveViewUnsupported` before the first `capture_preview()` call:

```
live view unavailable: ILCE-6000 does not support live view. Scanning still works, just without a preview.
```

Immediately after the refusal the session is still healthy — `iso=100`, `shutterspeed=1/13`, `f-number=f/4` all read back fine. I ran three open → refuse → close cycles and reconnected cleanly after each, plus again after a full app run. **No power cycle needed at any point**, which was the actual damage in the original report.

**Capture works on a session that refused live view.** Fired real frames on the same open session — 24.6 MB `.ARW` each, ~2.4 s per frame including download.

**And the captured file opens in NegPy.** Loaded one of those ARWs back into the app: decodes and renders through the full GPU pipeline (21 shaders, ~440 ms to first render), producing a normal `base_positive` buffer with real image content (mean 0.39, std 0.29 — not flat, no non-finite values). So the round trip a6000 → scan → open → process is intact end to end.

**UI behaves as described.** Drove the real app against the camera: the scan window opens with the preview pane replaced by the reason text, Scan and Retake stay enabled, the ISO/shutter/aperture steppers populate from the settings the refusal publishes on its way out (27/62/37 options), and Calibrate is greyed with the explanatory tooltip. The connection poll reports `camera_preview=False` / `camera_config=True` before live view is ever opened, so calibration is gated up front rather than after a failed attempt.

**Not covered by this test:** I don't currently have a Scanlight attached to this rig, so the RGB-triplet path and the calibration flow itself went untested, only the gating around them. I do own one, but don't have my camera scanning rig available to test for a few days. Will follow up when I can test that, but otherwise I think this PR fixes the root of the issue!

Thanks for the quick turnaround on this one.

PR #633, opened 2026-07-26
Fix macOS release builds for Intel + ARM runners

JKBusse:
## Summary
- split macOS runners by architecture to avoid cross-arch wheel mismatches
- use `macos-14` for arm64 and `macos-15-intel` for x86_64
- install `llvm@20` on macOS and export LLVM env for llvmlite in x86_64 jobs
- enable successful creation of Intel macOS DMG artifacts in release CI

## Why
Intel macOS packaging previously failed due to architecture-incompatible wheels and missing LLVM cmake config for llvmlite.

## Validation
- release workflow now produces an Intel DMG artifact successfully in CI
- arm64 macOS build path remains intact

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Install System Dependencies (macOS)
         if: startsWith(matrix.os, 'macos-')
-        run: brew install sane-backends
+        run: brew install sane-backends llvm@20
 
       - name: Install Packaging Tools (Windows)
         if: matrix.os == 'windows-latest'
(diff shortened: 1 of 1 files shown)


marcinz606:
nice work! I was holding that off because I don't have hardware to debug the intel mac builds

JKBusse:
Thanks, but Copilot did most of the work. I just verified and tested everything. If you ever want to test something on an Intel Mac, feel free to let me know!

JKBusse:
Ive tested the Intel-DMG from this Runs successfully: https://github.com/marcinz606/NegPy/actions/runs/30202746848

PR #641, opened 2026-07-27
Fix glow/halation luma to use Adobe RGB weights, not Rec.709

thetalkingdrum:
## Bug

`apply_glow_and_halation()` in `features/lab/logic.py` hardcoded Rec.709 luma
weights (0.2126/0.7152/0.0722) for both the glow highlight mask and the
halation mask, even though the buffer at this point is scene-linear Adobe RGB
(1998) — the same buffer `apply_rl_sharpening`, a few lines above in the same
file, already handles correctly via `LUM_R/LUM_G/LUM_B` (the Adobe RGB → XYZ
D65 Y row).

The GPU shader (`lab.wgsl`) had the identical split: a separate Rec.709
`LUMA_COEFFS` used only by the glow/halation taps, while the file's own
`rgb_to_lab()` already used the correct Adobe RGB Y row a few lines below.

## Fix

- CPU: both call sites now use `LUM_R/LUM_G/LUM_B`.
- GPU: renamed `LUMA_COEFFS` → `WORKING_LUMA_COEFFS` with the correct Adobe
  RGB values, keeping CPU/GPU in parity.

## Testing

- Full suite: 2493 passed, 6 skipped, 7 deselected, 0 failed. No golden
  values needed regeneration — `glow_amount`/`halation_strength` default to
  0.0 and the function no-ops at zero, so no default-config test exercises
  this path.
- Manually verified on a real scan (bus interior, mixed warm highlights)
  with glow=halation=0.5: max pixel diff 11/255, mean diff 0.05. Diff is
  concentrated almost entirely in bright, red-heavy highlight regions (mean
  brightness 149 vs. 45 image-wide, mean R value 176 vs. 60 image-wide) —
  consistent with R's weight increasing the most (0.213→0.297). Highlight
  clipping indicator moved from 1.7% to 1.8% at identical slider settings.
- Effect is real and correctly localized to highlights, but small — not
  something a user would likely notice by eye. This is a correctness fix
  (wrong luminance formula for the buffer's actual primaries) rather than a
  fix for a visible artifact.

---
Diagnosed and implemented with Claude Code; verified manually on a real scan
before opening this PR.

### DIFF
--- a/negpy/features/lab/logic.py
+++ b/negpy/features/lab/logic.py
@@ -281,7 +281,7 @@ def apply_glow_and_halation(
     if glow_amount > 0.0:
         # Highlight mask in the display domain (keeps the 0.5 threshold); bloom is linear.
         enc = working_oetf_encode(img)
-        luma = enc[:, :, 0] * 0.2126 + enc[:, :, 1] * 0.7152 + enc[:, :, 2] * 0.0722
+        luma = enc[:, :, 0] * np.float32(LUM_R) + enc[:, :, 1] * np.float32(LUM_G) + enc[:, :, 2] * np.float32(LUM_B)
         threshold = 0.5
         glow_mask = np.clip((luma - threshold) / (1.0 - threshold), 0.0, 1.0) ** 2
         base_r = max(3, int(15 * scale_factor))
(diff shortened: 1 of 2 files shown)


thetalkingdrum:
This is my first PR ever, so bear with me if I've missed any conventions,
happy to adjust anything.

I found this while going through the color pipeline end to end trying to
understand how it fits together, and this looked like a genuine (if small)
correctness bug rather than something I was specifically hunting for. Wanted
to submit it partly because it seemed worth fixing, and partly to get a feel
for how GitHub's PR process works.

PR #650, opened 2026-07-27
Disable sensor profile if linear raw is off

### LINKED ISSUE #649: Sensor Profiles only work correctly with Linear RAW enabled
The sensor profile feature does not work properly if linear raw is disabled, and can create nonsensical results. It is probably theoretically possible to make this feature work with camera wb, but for now it's probably best to just disallow the combination of sensor profile enabled and linear raw disabled.

seanharding:
## Skip the sensor unmix when Linear RAW is off

Sensor crosstalk matrices are always calibrated from neutral-WB decodes — the calibration dialog hard-codes `use_camera_wb=False`. But the correction was applied unconditionally, including when Linear RAW is off, where a RAW buffer instead carries the camera's as-shot per-channel gains.

Those gains don't commute with the unmix, so the matrix lands in the wrong basis. It doesn't just wash out — the residual comes back **sign-flipped**, so the correction overcorrects. On a representative narrowband response, a 22% green/blue leak was left at 13% with inverted terms. Silently wrong rather than silently weaker, which is why this is worth gating rather than tolerating.

### What this does

- Routes the three apply sites (`run_pipeline`, `_decode_oriented_f32`, and the batch-analysis worker) through `effective_sensor_matrix()`, which returns `None` in that basis.
- Folds the same gate into `sensor_token`, so two configs that render identically also hash identically.
- Greys the Sensor Calibration panel while gated: the dropdown shows `None`,  the calibrate button is disabled, and a muted "Requires Linear RAW." explains why. The reset is display-only — `sensor_profile` is untouched, so the selection survives a Linear RAW round-trip.

### Notes
  
The math itself is basis-agnostic: column normalization makes the matrix transform by similarity, so calibrating in the *same* WB basis would be exact. Lifting the gate later means recording the basis in the profile TOML and requiring a locked manual WB across the three bare-light shots. There's a test covering this so the path stays documented.

Tests added for the gate, the basis mismatch, and the panel state. Full suite green.

Fixes #649

### DIFF
--- a/negpy/desktop/view/sidebar/sensor.py
+++ b/negpy/desktop/view/sidebar/sensor.py
@@ -1,7 +1,7 @@
 from PyQt6.QtWidgets import QComboBox, QHBoxLayout
 
 from negpy.desktop.view.sidebar.base import BaseSidebar
-from negpy.desktop.view.styles.templates import field_label, hint_label
+from negpy.desktop.view.styles.templates import field_label, hint_label, wrap_tooltip
 from negpy.features.process.models import invalidate_local_bounds
 from negpy.features.process.sensor import sensor_unmix_available
 from negpy.services.assets.sensor import SensorProfiles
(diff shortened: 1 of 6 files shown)


thetalkingdrum:
Genuine question (I don't even own a digital camera so tell me if I'm off): if Linear RAW off is just using use_camera_wb=True under the hood, isn't libraw already reading the as-shot WB multipliers from the file to do that? If so, could the fix be simpler than gating it off, just grab that same multiplier at decode time, divide it back out before the matrix, apply the matrix as normal, done? Rather than requiring calibration to happen under one locked WB going forward.

Might be missing something about where that value is or isn't accessible in the pipeline, but seemed worth floating in case the fix is closer than it looks.

seanharding:
Hmm, yeah, valid question. I'll need to take some time to think it through and test it to make sure it works as expected. At a gut level, it feels to me like applying a correction matrix based on one set of white balance multipliers, and then applying a different set for the rest of the processing, might still sometimes cause unpredictable results, but maybe I'm overthinking it...

thetalkingdrum:
Or could we just fold the two together instead- Combine the WB gain and the correction matrix into one, then apply that directly to the as-shot data? No separate un-scale step needed. Might even fit effective_sensor_matrix() nicely — return that combined matrix instead of None when the gain's known.

thetalkingdrum:
But yeah, just got the epiphany when I saw this. So maybe there's something I'm missing.

thetalkingdrum:
Yeah, your gut feeling was right.  Claude didn't like my original idea either. I think you're right that just doing M · G⁻¹ would leave things in the wrong basis for the rest of the pipeline.

But might still be fixable on the fly: instead of dividing out the WB gain, correcting, and stopping there, what if you reapply the gain after — so G · M · G⁻¹ instead of just M · G⁻¹? That keeps the output back in the same as-shot WB basis everything downstream expects, rather than handing it native-basis values. Still just matrices you already have, still computed per-decode, no locked calibration WB needed.

marcinz606:
I think this PR change makes sense, once calibration is ran then as-shot multipliers should be thrown away

PR #615, opened 2026-07-23
feat(roll): add coolscanpy roll-scanning backend adapter (optional dependency)

rohanpandula:
## Summary

This is the thin connector from the standalone [`coolscanpy`](https://github.com/rohanpandula/coolscanpy) package into NegPy — the follow-through on the agreement in #497. coolscanpy owns all the device-specific LS-5000/SA-21 USB transport; NegPy consumes it exactly the way it already consumes `python-sane` (plain Scan panel) and `gphoto2` (Camera Scanning): as an **optional dependency, imported lazily, entirely absent by default**.

It is intentionally scoped to *only* the glue. No desktop sidebar, no background worker, no repair engine, no output tiers, no Nikon-exact color builders, no live-evidence contracts — all of that is meant to land as separate review units stacked on top of this one. This PR adds nothing a reviewer has to maintain beyond a small adapter and the file-writing service behind it.

## What it adds

- **`coolscan-roll` optional dependency group** in `pyproject.toml`, mirroring the existing `scanner` (python-sane) and `camera` (gphoto2) groups. Pins coolscanpy to immutable commit `970d18e`. `uv sync --group coolscan-roll` opts in; the rest of the app never sees it.
- **`negpy/infrastructure/roll/coolscanpy_roll.py`** — the only module that imports coolscanpy. Wraps `coolscanpy.Device` + `Roll` behind a small `RollHandle` (preview, spacing correction, approve, batch `scan_many`, `safe_stop`, `close`), and translates coolscanpy's typed `PyCoolscanError` hierarchy to plain `RuntimeError`s the same way `SaneBackend` already reports failures — original preserved as `__cause__`. `available()` is a cheap `importlib.util.find_spec()` check, mirroring `ScanSidebar._sane_available()`, so the feature degrades to fully absent when coolscanpy isn't installed.
- **`negpy/services/roll/service.py`** — `RollScanningService` orchestrates open/close (one reservation at a time, matching coolscanpy's own lock), preview, approve, spacing correction, batch `scan_many`, and `safe_stop`. `write_frame()` writes each frame in NegPy's existing scan-output shape: a 16-bit TIFF master, an `_IR.tif` sidecar when the frame carries infrared (matching `writer.write_tiff_16bit`'s own suffix), and a `_receipt.json` sidecar. The filename pattern is the same Jinja2 template the plain Scan panel uses, but `seq` is seeded from the frame's physical slot number so re-scanning one bad slot overwrites instead of accumulating a second copy.
[...]

PR #652, opened 2026-07-27
feat(capture): add Fujifilm Bluetooth trigger connector (optional fuji-ble backend)

rohanpandula:
## Summary

This is the thin connector from the standalone [`fuji-ble-negpy-trigger`](https://github.com/rohanpandula/fuji-ble-negpy-trigger) package into NegPy — the same shape as #615 (coolscanpy) and the way NegPy already consumes `python-sane` (plain Scan panel) and `gphoto2` (Camera Scanning): an **optional dependency, imported lazily, entirely absent by default**.

It lets a Fujifilm GFX body be triggered **over Bluetooth LE** for automated RGB film scanning. The reason it exists: Fujifilm bodies stick in a gphoto2 tethered-capture state and refuse to reconnect, so gphoto2 *triggering* is unusable on them. Here the shutter fires over BLE and the RAW arrives via the camera's WiFi "PC Auto Save" into a watched folder — no gphoto2 capture session, so nothing gets stuck.

It is intentionally scoped to the glue. **No sidebar UI in this PR** — that lands stacked on top (#follow-up). This PR adds nothing a reviewer has to maintain beyond a small adapter, a backend seam in the worker, and the tests behind them.

## What it adds

- **`fuji-ble` optional dependency group** in `pyproject.toml`, mirroring the existing `scanner` (python-sane) and `camera` (gphoto2) groups. Pins `fuji-ble-negpy-trigger` to an immutable commit, exactly as the coolscan-roll group pins coolscanpy. `uv sync --group fuji-ble` opts in; the rest of the app never sees it. macOS/Linux only (bleak/CoreBluetooth).
- **`negpy/infrastructure/capture/fuji_ble.py`** — the **only** module that imports `fujitrigger`. Exposes `available()` (a cheap `importlib.util.find_spec` check, mirroring `ScanSidebar._sane_available()`), `make_camera()`, `discover()`, and `pair()`. It accepts and returns **plain dicts** — no `fujitrigger` type escapes this module — and translates vendor errors to plain `RuntimeError` with the original preserved as `__cause__`, the same contract `SaneBackend` and `GphotoCamera` already honour.
- **`Camera` protocol gains `is_open()`** (`base.py`) so the worker treats every backend uniformly — gphoto2 USB and the Bluetooth trigger both report it, and the worker never calls a backend-specific method to ask. This is the seam: a backend is selected and constructed through the adapter, not special-cased.
- **`CaptureWorker` backend seam** — `set_camera_backend()` arms `usb` (gphoto2, unchanged) or `fuji_ble`; the BLE camera is built through `fuji_ble.make_camera()` and live-view/settings/calibration gate themselves off via capabilities (the remote carries only the shutter). The worker stores the Fuji config as an **opaque dict** it never interprets.
- **`build.py`** — bundles `fujitrigger` + `bleak` (both imported lazily, so PyInstaller would miss them) and adds `NSBluetoothAlwaysUsageDescription` to the macOS app's Info.plist when the group is installed, mirroring the gphoto2 optional-extra handling.
- **`tests/test_capture_fuji_ble.py`** — injects a **fake `fujitrigger` module**, so no Bluetooth hardware and no real package are needed (the same approach #615 uses for coolscanpy).

## The integration point

`fuji_ble.py` is the one place `fujitrigger` is imported. Everything above it — the worker, and the sidebar in the follow-up — talks only through its functions and plain dicts. Re-pointing the device code later means changing only this module.

## Hardware & validation

Validated against one setup: **Fujifilm GFX 100 II on macOS** — paired over BLE (CoreBluetooth-managed SMP bond), completed the secure handshake, fired the shutter, and reconnected by serial across the body's rotating pre-bond addresses. The protocol is ported from the [`furble`](https://github.com/gkoh/furble) firmware (MIT). Every other GFX/X body is untested.

## Verification

- `tests/test_capture_fuji_ble.py` + `tests/test_capture_worker.py` pass against the fake-`fujitrigger` fixture with no hardware (and the full suite: 2545 passed).
- `make lint` and `make type` pass.
- Feature is fully inert when `fuji-ble-negpy-trigger` is absent — `available()` gates every entry point.

## Review note

Opened without the sidebar so the boundary shape can be reviewed on its own. The pairing UI (Trigger selector, Pair button, drop-folder picker) is ready as a stacked follow-up. Feedback on the adapter boundary and the `is_open()`/backend seam is what I'm most looking for.

### DIFF
--- a/build.py
+++ b/build.py
@@ -114,6 +114,30 @@ def collect_gphoto2_plugins() -> None:
 
 collect_gphoto2_plugins()
 
+
+def collect_fuji_ble() -> None:
+    """Ship the optional Fujifilm Bluetooth trigger (fuji-ble-negpy-trigger + bleak).
+
+    Both are imported lazily by ``negpy.infrastructure.capture.fuji_ble``, so PyInstaller
+    would miss them — and bleak's CoreBluetooth backend (pyobjc) must be bundled for
+    Bluetooth to work in the packaged app. Mirrors the gphoto2 optional-extra pattern:
+    ``uv sync --group fuji-ble`` before building, or the packaged app simply shows its
+    setup hint.
+    """
+    if is_windows:
+        return  # the Bluetooth trigger is macOS/Linux only
+    try:
+        import fujitrigger  # noqa: F401
+    except ImportError:
+        print("fuji-ble-negpy-trigger not installed — packaging without the Bluetooth trigger")
+        return
+    params.append("--collect-all=fujitrigger")
+    params.append("--collect-all=bleak")
+    print("Bundling Fujifilm Bluetooth trigger (fujitrigger + bleak)")
+
+
+collect_fuji_ble()
+
 # Add platform-specific icon
 if is_windows:
     icon_path = os.path.abspath("media/icons/icon.ico")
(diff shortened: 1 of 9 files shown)


light-sntchr:
Is this still necessary after the latest fixes?

PR #664, opened 2026-07-29
Add density-space Print Saturation, composed into the paper's dye crosstalk matrix

thetalkingdrum:
Adds a Saturation control that operates in density space — composed into the exact same matrix slot as the paper profile's dye crosstalk (dye_matrix) — as an alternative to scaling Lab a*/b* post-decode. New field ExposureConfig.density_saturation (+ per-channel trims), default 1.0 = identity.

Lab Saturation scales a*/b* uniformly around L* — a purely perceptual-space operation with no relationship to how a physical print produces color; it's a math trick applied after decode. This instead pushes/pulls the print's actual per-channel dye densities apart, in the same coordinate space the paper's own dye crosstalk already operates in, composed via sat_matrix @ dye_matrix into one matmul.

Two consequences follow directly from that, not as separate design choices:
- It composes correctly with different paper profiles. Because it's literally the same matrix slot as the profile's dye crosstalk, pushing separation on a paper with real (impure) dye response looks different than on a cleaner profile — the way it would on an actual darkroom print — for free, with no per-paper special-casing.
- Its effect naturally scales with how much density-above-base a channel already has. Saturation is applied to density - paper_base after the full per-channel H&D curve (toe/shoulder/grade/etc.) has already run — so a channel sitting near its shoulder, already compressed by the curve, starts from a smaller density-above-base value than one in the linear region, and gets a proportionally smaller absolute push.

It also supports independent per-channel (R/G/B) trims, matching the existing pattern already used elsewhere in this section (Grade trim, Toe/Shoulder trim, Snap/midtone-gamma trim all support per-layer correction) — none of which the Lab sidebar's controls (Saturation, Vibrance, Dye Mute) currently support at all. That's not incidental to where this control lives: a real paper's dye layers can diverge independently (crossover), and this control can now correct that the same way the other Tone-stage controls already do. Placing it here keeps it consistent with that established convention rather than introducing a first per-channel-capable control in a section that has no precedent for one.

New field, defaults to identity (1.0). compose_density_matrices returns the dye matrix unmodified at default, so output is bit-for-bit unchanged both for the neutral paper (test_identity_is_exact_noop) and for papers with real dye crosstalk like Kodak Endura
(test_identity_is_exact_noop_with_real_dye_matrix). Cannot affect any existing saved edit.

Implemented on both the CPU and GPU render paths — verified by test_cpu_gpu_match_density_saturation (tests/test_gpu_curve_parity.py), which renders through the real GPU compute path with a non-identity paper dye matrix and non-default saturation/trims. No CPU-only fallback or silent no-op under the app's default GPU-first rendering.

- tests/test_density_saturation.py — matrix math (row-stochastic under per-channel divergence, sat @ dye composition order), per-channel resolve, and end-to-end render behavior (exact no-op at identity for both neutral and non-identity-dye papers, real effect, per-channel trims, B&W inertness)
- tests/test_gpu_curve_parity.py::test_cpu_gpu_match_density_saturation — CPU/GPU parity against a real paper dye matrix
- docs/PIPELINE.md updated

### DIFF
--- a/docs/PIPELINE.md
+++ b/docs/PIPELINE.md
@@ -104,6 +104,11 @@ A **paper profile** (`paper_profile`, default *Neutral*) overrides the print *ch
 
 Profiles are **mode-aware**: C-41 exposes the RA4 colour papers, B&W exposes the tonal-only B&W papers (paper tone is a Toning job, so B&W profiles carry no colour terms), and E-6 gets only *Neutral*. An incompatible stored value collapses to *Neutral* so it can never leak into a render. Bundled papers: **Neutral**; *B&W* — Ilford Multigrade RC, Ilford Multigrade FB Classic, Foma Fomatone, Foma Fomabrom; *RA4* — Kodak Endura Premier, Fujicolor Crystal Archive. Values are loosely mapped from datasheets (mainly $D_{max}$ is grounded; the knee/midtone tweaks are light character touches).
 
+### Print Saturation
+**Code**: `negpy.features.exposure.papers.resolve_saturation_matrix` / `compose_density_matrices`
+
+Saturation applied in the same density domain and the same matrix slot as a paper's `dye_matrix`, rather than as a post-decode CIELAB $a^{\ast}/b^{\ast}$ scale (contrast with §6's Global Saturation). For density above paper base $e = D - D_{min}$, the saturation matrix is $M(k) = k \cdot I + (1-k) \cdot J$ ($J$ = all-rows-$\tfrac{1}{3}$, the achromatic projection): $k=1$ identity, $k=0$ collapses that channel's row to the achromatic mean, $k>1$ boosts density separation. Composed as $M_{total} = M(k) \cdot M_{dye}$ — saturation outermost, the paper's own dye coupling innermost, so `dye_matrix` keeps acting on the print curve's real density unchanged and saturation layers on top as a final step rather than being partly reabsorbed by the paper's crosstalk. $k$ takes independent per-channel trims (`density_saturation_trim_red/green/blue`) the same way Grade/Toe/Shoulder do; each row still sums to 1 regardless of its own $k$, so neutrals stay flat at any per-channel divergence.
+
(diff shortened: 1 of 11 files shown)


marcinz606:
really nice idea and results look really good. I'll do one more thing as follow up - move dye mute to this stage as modifier to density saturation

thetalkingdrum:
Thanks! Yeah it’s something I have been thinking about for some time.

I actually made a new dye mute that works slightly differently to go with
the density saturation. But I wasn’t completely done yet so I just pushed
this one first to get it out the door.

ons. 29. juli 2026 kl. 07:25 skrev marcinz606 ***@***.***>:

marcinz606:
i already did the changing the dye mute to work with it here: https://github.com/marcinz606/NegPy/pull/667 but maybe your idea was slightly different?

thetalkingdrum:
yeah I've been working on a density-based vibrance/anti-vibrance concept,
and my dye mute work borrows from that anti-vibrance side of it. None of
it's fully done yet, but I think it's giving some nice real-world results.
I can put together a draft soon if you want to try it yourself? Then you
can see if it something you are interesed in going further with. The
density-based vibrance looks pretty similar to normal Lab vibrance, hard to
say which is actually better there. But the "anti-vibrance" pairs really
well with density saturation for taming already-saturated colors. Main
difference from your slope-based damping is that it's per-pixel rather than
frame-wide, so it only tames pixels that are actually already saturated
instead of damping everything in the frame the same amount.

marcinz606:
do the draft PR replacing the dye mute with your solution please, I'll check it out then. Sounds interesting

marcinz606:
I've also made the dye mute default 0.0 after moving it to density space

thetalkingdrum:
Ok. Made the draft. Left in my own A/B test tools and the separate density
vibrance slider so you can get a feel for the whole concept. I can remove that
stuff if that works better. Note: vibrance turns to anti-vibrance at negative
values

PR #675, opened 2026-07-29
Draft: density-space Vibrance/anti-vibrance + spatially-modulated Dye Mute (extends #667)

thetalkingdrum:
# Draft: density-space Vibrance/anti-vibrance + spatially-modulated Dye Mute (extends #667)

## What

Two related, additive controls, both off by default:

- **Density Vibrance** — a signed per-pixel Vibrance/anti-vibrance control in density space (`ExposureConfig.density_vibrance`).
- **Spatial Damping** — an alternate, per-pixel application mode for Dye Mute (`ExposureConfig.density_damping_spatial`), using the same masking mechanism as anti-vibrance.

This is a **draft** — I'm looking for your read on whether the direction is worth continuing, not asking you to merge as-is.  I already had some A/B testing built in that I decided to keep as it might make your testing easier. Happy to close and keep iterating if it doesn't look right, or rip out some of the A/B features. 

## Why

`#667` moved Dye Mute into density space as a single frame-wide scalar, grade-coupled via `slope_g`. That fixed the domain problem (density vs. post-decode Lab), but the correction is still uniform — it can't distinguish a near-neutral pixel from an already-separated one.

Anti-vibrance needed the same distinction (compress already-vivid pixels, leave muted ones alone), so one per-pixel masking mechanism ended up answering both:
- Applied on its own — **Density Vibrance**, sign-flippable, boost muted or compress vivid.
- Applied to Dye Mute's strength instead of the Vibrance slider's — **Spatial Damping**, the same correction redistributed by how separated each pixel already is, instead of applied uniformly.

Same math, two applications — not three separate ideas.

## What each control does

**Dye Mute** — unchanged: `#667` exactly as shipped, one flat `(slope_min/slope_g)^strength` damp applied uniformly, grade-coupled. This PR doesn't touch its default behavior.

**Spatial Damping** (new toggle, off by default) — Dye Mute run through anti-vibrance's math instead of the uniform formula. Same strength value, same field, redistributed per-pixel via `d_k = 1 + (damp_global - 1) * mask` — the same shape as the anti-vibrance formula below, just fed by Dye Mute's strength.
- **Off** — byte-identical to `#667` (verified, see Testing).
- **On** — same strength, scaled by how separated each pixel already is: near-neutral pixels get none of it, already-separated pixels get up to full strength. Replaces the uniform formula rather than layering on it — see "Replace vs. layer" below.

Doesn't touch the Vibrance slider or Print Saturation — a second application of the masking trick, not a shared control.

**Density Vibrance** (new slider, `-1.0` to `1.0`, `0` = off):
- **Positive** — classic Vibrance: boosts muted pixels (mask targets low-spread pixels).
- **Negative** — real anti-vibrance: compresses vivid pixels (mask targets high-spread pixels — the same mask Spatial Damping reuses), leaves muted pixels alone. Lab's Vibrance has no equivalent here — its sub-1.0 branch just desaturates the muted population further.

The mask targets a genuinely different population depending on sign — not just a negated output.

**A/B row** (`Off / Density +0.5 / Lab 1.50 / Density -0.3 / Lab 0.33`) — left over from my own testing against Lab Vibrance, not built specifically for review, but there if useful. The boost pair is a fair comparison (both target muted pixels, matched by chroma effect). The compress pair is magnitude-matched only — Lab's 0.33 desaturates muted pixels further, Density's -0.3 compresses vivid pixels, so they'll look different even at "equal" strength.

**Shared constant to flag:** all three per-pixel behaviors — Vibrance boost, anti-vibrance, Spatial Damping — share one tunable, `vibrance_spread_scale = 0.4` (`exposure/models.py`), controlling how much per-pixel separation counts as "vivid enough" to affect. It's an unvalidated starting guess. Tuning it for one effect will retune all three at once — that's the main risk of the unified-mask approach, and worth deciding now whether that coupling is acceptable or whether it needs decoupling before this ships.

## Honest note on the boost direction

Density Vibrance's boost direction converged with Lab Vibrance at realistic settings in my testing, only diverging at extremes — I can't confidently say one is better. Including it anyway since it costs nothing and the A/B row makes it checkable, but the real case for this PR is the compress direction, which Lab has no equivalent of at all.

## Replace vs. layer

Spatial Damping replaces the uniform `grade_saturation_damping` pre-multiply when it's on, rather than layering on top. I lean toward replace for three reasons: your own `#667` PR already diagnosed the failure mode of uncoordinated stacking (Lab Saturation vs. old Dye Mute fighting each other); and two of my own earlier prototypes (a global chroma-damping signal, then a per-channel-divergence version — both dropped, see `DROPPED_KEYS`) turned out mathematically redundant with `density_saturation` once measured. Flagging this as a decision point rather than deciding it unilaterally — layer is a reasonable position if you'd rather.

## Risk / compatibility

Both fields default to off/identity (`density_vibrance=0.0`, `density_damping_spatial=False`) — can't affect any existing saved edit. With Spatial Damping off, Dye Mute is unchanged from `#667`, verified bit-exact (`test_off_by_default_matches_uniform_damping_bit_exact`).

Implemented on CPU and GPU, mirrored in the WGSL shader — no CPU-only fallback.

## Testing

- `tests/test_density_vibrance.py` — identity/no-op, boost and compress each change output and aren't a simple sign flip of each other, B&W inertness, bit-exact proof Spatial Damping off matches `#667`'s uniform path.
- `tests/test_gpu_curve_parity.py` — new CPU/GPU parity for Density Vibrance (both directions) and Spatial Damping — neither covered by the existing `test_cpu_gpu_match_dye_mute`, which only exercises the uniform path.

### DIFF
--- a/negpy/desktop/view/sidebar/tone.py
+++ b/negpy/desktop/view/sidebar/tone.py
@@ -11,6 +11,20 @@ _CH_SUFFIX = ("red", "green", "blue")
 _CH_LABEL = ("", " R", " G", " B")
 _CH_COLORS = ("#ff5a5a", "#5adc78", "#5f96ff")
 
+# PROTOTYPE A/B strengths for Density Vibrance vs Lab Vibrance, matched by
+# overall mean-chroma effect on a real test frame (07.raw, neutral paper) —
+# see the density-saturation investigation memory for the render scan.
+# Boost direction (+0.5 density / 1.50 Lab) is a fair comparison: both target
+# the same population (muted pixels). Compress direction (-0.5 density /
+# 0.33 Lab) is magnitude-matched only -- Lab's sub-1.0 desaturates MUTED
+# pixels further, Density's negative compresses VIVID pixels; they'll look
+# different even at "equal" overall chroma change, not just differently
+# strong. Not a bug, worth seeing directly in the A/B.
+_VIBRANCE_AB_DENSITY_BOOST = 0.5
+_VIBRANCE_AB_LAB_BOOST = 1.50
+_VIBRANCE_AB_DENSITY_COMPRESS = -0.5
+_VIBRANCE_AB_LAB_COMPRESS = 0.33
+
 
 class ToneSidebar(BaseSidebar):
     """Print/zone density, Grade, Print Saturation, paper white, and a labeled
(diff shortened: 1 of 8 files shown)


marcinz606:
I played around and I like the effect, i think it should replace the dye mute. But for consistency/simplification I would maybe name it "Dye Separation" instead of "density vibrance". I can make a changes on top of your branch

marcinz606:
Also I think i will drop vibrance from lab panel completely after that

thetalkingdrum:
Stoked to hear that you will include it :) Yeah, after having played around with the density vibrance and density saturation in tandem the last few days I also really started liking it. feels like a quite powerful but subtle tool for getting the saturation you want. And looks natural.

thetalkingdrum:
No objections on calling it Dye separation:)

thetalkingdrum:
So go ahead and make the chances you want. The only thing that could be explored more imo is if some other value than 0.4 gives a better result. But that can be done over a time period as well. Don't need to land that right away.

marcinz606:
created new branch/pr on top of your commits here #676, i will close this one :)

marcinz606:
I'm playing with spread scale and i think smaller values work better, with current 0.4 behaviour is pretty close to print saturation

marcinz606:
something in 0.1-0.2 range seems way more selective

thetalkingdrum:
I was thinking something a bit smaller might be better too:)

ons. 29. juli 2026 kl. 20:14 skrev marcinz606 ***@***.***>:

thetalkingdrum:
0.4 seemed a bit heavy handed

ons. 29. juli 2026 kl. 20:17 skrev Mats Fagerberg ***@***.***>:

thetalkingdrum:
i synced with main now. Dye separation seems make almost no visible change between 0 and 1? Going the other direction works well though. 0 to -1

thetalkingdrum:
like the histogram changes but i barely see any visual change

marcinz606:
i reduced the k 0.4 -> 0.12, might need some more tweaking

marcinz606:
now the positive effect more heavily depends on negative

marcinz606:
but I agree it's weird that negative effect is more pronounced, i havent changed anything other than the 0.4 const

thetalkingdrum:
I'm heading out for a bit so will look more later. But yeah, on the test negative going in the positive direction made almost no difference

thetalkingdrum:
might be possible to use different k number on the two directions as well, if that works better idk?

marcinz606:
Yes, that's what I'm considering, different k depending on direction

marcinz606:
After playing a bit more I'm thinking that when effect is made stronger again it's almost indistinguishable from print saturation, formulas are almost the same after all. I'm considering consolidating it into one slider to make things simpler

thetalkingdrum:
Yeah, it’s quite subtle. Or that was my original idea. Something that would
let you push or pull saturated colors in a way that would blend in and look
natural. But if you decide too only keep one tool I think the print
saturation is the purer tool.

ons. 29. juli 2026 kl. 21:54 skrev marcinz606 ***@***.***>:

thetalkingdrum:
like if you want to push the saturation on an image with print sat, but
then there is one strong color like red that goes too far. Then you can get
that red back in line with the rest by compensating with the anti-vibrance.
That is how I was using it under testing the last few days.

ons. 29. juli 2026 kl. 22:12 skrev Mats Fagerberg ***@***.***>:

marcinz606:
I've made a hybrid now, one 'Dye Saturation' slider but you can dial in the "k" using separate slider. Gives a lot of flexibility

thetalkingdrum:
Yes, I just noticed:) was looking over the docs now. Will be interesting to
try out.

thetalkingdrum:
Seems to work well and is flexible as you say. It reacts a bit differently on different frames, so I have to recalibrate my brain:)

thetalkingdrum:
I'm convinced. Gives you very fine saturation control over the whole image.

thetalkingdrum:
Since Chroma is still in the lab panel I made some changes to it so it behaves a bit better.

thetalkingdrum:
Test with some skin tones while boosting chroma for best a/b comparison

PR #695, opened 2026-07-30
Make Lab Chroma gamut-aware, add skin-tone protection

thetalkingdrum:
## What
Makes Lab **Chroma** (above 1.0) gamut-aware instead of hard-clamping, and adds a boost-only skin-tone protection band on top of it.

## Why
The old flat `a*/b*` scale mathematically preserves hue — uniform scaling of both chroma components leaves `atan2(b, a)` unchanged — but pixels that overshot the display gamut then got clamped per RGB channel afterward. Clamping only the channel(s) that overshot changes the R:G:B ratio the eye actually sees, so the *visible* hue shifted even though the math didn't intend it to. Most visible pushing saturated frames hard: reds go orange, greens go yellow.

This replaces the hard clamp with a per-pixel soft knee: pixels comfortably inside gamut at full strength get the same flat scale as before (byte-identical below 1.0, unchanged), pixels that would clip get bisected toward their own actual in-gamut headroom and eased in with a softplus-style knee — the same knee shape the print curve already uses for toe/shoulder. CPU (`gamut_aware_chroma_scale` in `kernel/image/logic.py`) and GPU (`lab.wgsl`) mirror each other exactly; parity covered in `tests/test_pipeline_parity.py`.

On top of that: pushing Chroma up disproportionately affects skin tones, since faces sit in a fairly narrow, moderate-chroma hue band that saturates faster than most background colour. `_skin_protection_weight` softens the boost (never the cut) in a Gaussian band centered at hue ≈ 52°, width 25°, measured from a rendered portrait frame's actual skin-pixel hue distribution — a rough single-frame mask, not a validated skin detector, so treat the constants as a starting point worth checking against more frames. Not exposed as a control — a flat, always-on improvement to the existing Chroma slider.

## Review
Went through a draft with an in-app A/B toggle so the old and new behavior could be compared directly on the same Chroma value. @marcinz606 A/B-checked it against several frames and preferred the gamut-aware/skin-protected result on all of them, so the toggle has been removed — this is now the only behavior, no new control added.

## Testing
- `make all` (lint + type + full suite) passes clean.
- New/extended coverage in `tests/test_lab_logic.py` and `tests/test_pipeline_parity.py`.
- Verified interactively in-app against real C-41 frames, including the A/B comparison above.
- Docs updated in the same change: `docs/PIPELINE.md` §6.2 and `docs/USER_GUIDE.md`.


marcinz606:
I rarely add saturation but I did some a/b checks on various frames and on all I prefered "new" look. I'd say go ahead, remove the toggle and just replace it completely with new behaviour :)

thetalkingdrum:
Yes, I don't use it either, but I thought since it stays it can at least be the best possible version. If anyone uses it:)

thetalkingdrum:
done!

PR #702, opened 2026-07-31
Skin Protection as a Lab control, working at any Chroma

marcinz606:
## Why

The skin-tone protection added in #695 is welded into the Chroma boost path: it damps the *requested saturation* in a skin-hue band, boost-only, at a hardcoded strength. Two consequences:

- **Inert at Chroma 1.0.** `apply_saturation` early-returns there and the processor never calls it, so a face that came off the print curve already sunburnt got nothing — the tool only guarded against its own push.
- **Hue-only mask.** In this working space that cannot tell a face from a red car: pure saturated red sits at ~40°, well inside the 52°±25° Gaussian, and got damped like skin. The #695 commit message flagged this as accepted-for-now.

## What

One operator replaces the boost-damping term: a **soft chroma ceiling inside a skin mask**, applied after the saturation scale and independent of it, exposed as a **Skin Protection** slider in the Lab panel (0–1, default 0).

A ceiling subsumes the damping — pushing skin past it gets reined in, which is what the damping was for — while also catching skin that arrived over-chromatic from the print curve. It is monotone and one-directional (chroma is only ever reduced, never added), so `saturation = 0.0` still reaches true grey without the boost-only special case the old band needed, and it is safe to run after the gamut knee, since a smaller chroma cannot overshoot a gamut the full-strength push already fitted inside.

**Mask** (`_skin_weight`) — axis-aligned reduction of the CIELAB skin locus, a product of three terms:

| axis | term | why |
|---|---|---|
| hue | Gaussian, 52° ± 25° | measured in #695; the literature places skin at 40–65° across the Monk/Fitzpatrick range |
| chroma | full to C\* 55, zero by 85 | **the new selectivity** — skin ≈ 27, sunburn ≈ 60, pure red ≈ 104 |
| lightness | rolloff below L\* 15, above 95 | hue angle is noise at both ends |

**Operator** — `ceiling = 22 / strength`, knee at `0.6 × ceiling`, same softplus shape as the gamut knee and the print curve's toe/shoulder. `a*` and `b*` scale together, so hue and L\* are untouched. The ceiling is a reciprocal rather than a lerp to some finite "off" value so it runs off past any reachable chroma as strength → 0 and the control fades out continuously instead of stepping the moment the slider leaves 0.

Prior art surveyed: classic skin detectors (Jones & Rehg, Hsu, Kovač) all use a 2-D chrominance region rather than hue alone; Capture One's Skin Tone tool adds a hue/sat/lightness *uniformity* pull toward the target. That hue-uniformity half is deliberately **not** built here — hue-warping is far more visible when the mask is wrong, and it rides the same mask if we want it later.

## Default is off

Zero drift on existing edits, and no always-on operator at Chroma 1.0 where nothing ran before. The trade-off, stated plainly: an existing edit at Chroma 1.6 renders slightly more saturated on skin than it did, until the slider is dialled in.

## Verification

- `make all` green; full suite 2893 passed. Goldens unchanged, as they must be with the operator skipped at 0.
- New CPU/GPU parity case at Chroma 1.0 + protection 0.8 — the path the saturation scale never reaches, so it is the only cover for the WGSL mirror. Uses colour patches rather than a gradient (per-channel normalization flattens a uniform cast); half the frame lands at mask weight > 0.3 and protection alone moves it by 0.11, so a mirror that silently no-ops fails.
- Real Ektar frame through the running app (headless driver): warm-band mean C\* **36.1 → 34.6 → 27.1** across 0 / 0.5 / 1.0, while the 352k-pixel non-warm control held at **50.056 → 50.055**. At engine level the hue delta is **0.00000° median / 0.00006° max**.

Docs updated in `USER_GUIDE.md` (panel:lab) and `PIPELINE.md` §6.


thetalkingdrum:
Nice idea! I only wanted to make Chroma "less bad" as it was basically unusable if any skin tone was in the frame. But this is more flexible and works even if the source is not the chroma slider. 

I would keep the gamut aware "soft knee" in Chroma though, as it actually did more alone than the skin tone band did.

marcinz606:
soft knee is in chroma only, I might do more tweaking to skin thing though, seems to affect a lot of reds/oranges a bit too much. It needs some more narrowing IMO

thetalkingdrum:
yeah, i meant if you take out skin protection from chroma, leave the soft knee in chroma. But the skin protection I left there was a pretty crude approximate value. 

But I'm unsure if chroma should stay long term..I'm not sure how useful it is in the pipeline. I just tried to make it a bit better.

PR #712, opened 2026-07-31
Fix test writing junk files to the real export directory

thetalkingdrum:
Had some ghost jpg files appear in my export folder. Took a while to figure out. This is the fix. 

test_batch_progress used DEFAULT_WORKSPACE_CONFIG.export which resolves to the user's actual export folder. Every test run wrote three 4-byte "bits" files (a.jpg, a.jpg, b.jpg) there. Override output_path to tmp_path so the test stays sandboxed.

Also remove the debug file-write tracing from main.py that was added to hunt this down.

### DIFF
--- a/tests/test_batch_progress.py
+++ b/tests/test_batch_progress.py
@@ -1,3 +1,4 @@
+from dataclasses import replace
 from unittest.mock import MagicMock
 
 from PyQt6.QtCore import Qt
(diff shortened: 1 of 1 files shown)


seanharding:
Haha, thank you! This has been bugging me too but I didn't get around to looking at it yet.

thetalkingdrum:
Haha, yes it was bugging me too.

PR #713, opened 2026-07-31
Disable LibRaw adjust_maximum_thr in _decode_sensor_rgb (?)

thetalkingdrum:
## Summary
`_decode_sensor_rgb` uses LibRaw's default `adjust_maximum_thr=0.75`, which silently substitutes the camera's nominal white level with the frame's own brightest pixel when it exceeds 75% of the stated maximum. This makes the scaling reference image-dependent — a bright specular in one frame shifts the decode of the entire tonal range relative to a frame without one.

Setting `adjust_maximum_thr=0.0` pins the reference to the camera's fixed white level. The calibration path (`raw_demosaic.py`) already carries this fix with documentation explaining the rationale; the main pipeline is the remaining gap.

## Why this matters
When `adjust_maximum_thr` is left at the default, LibRaw picks a different white-level reference per frame based on that frame's content. This means two frames from the same roll, shot with the same camera, can decode to different absolute values for the same scene luminance — the decode is no longer a fixed function of the sensor data.

This has knock-on effects across NegPy:
- **Batch normalization drifts.** The floor/ceiling analysis assumes all frames in a roll share the same decode scaling. A single bright frame that triggers the substitution will have its tonal range compressed relative to the others, skewing the batch statistics.
- **A/B consistency breaks.** Switching between two frames that were decoded with different effective white levels introduces a tonal shift that has nothing to do with exposure or scene content.
- **Roll-level operations become scene-dependent.** Crosstalk correction, flatfield, and white-balance normalization all assume the input is scaled to a fixed reference. A per-frame rescale means these corrections are applied to data that isn't on the same footing.
- **Non-deterministic re-decode.** Reopening the same file after editing a bright region (e.g. cropping out a specular) can change whether the threshold triggers, producing a different decode from the same source.

The calibration path (`raw_demosaic.py`) already documents this and uses `0.0`. The threshold exists in LibRaw to handle sensors where the nominal white level in the EXIF is too conservative, but for NegPy's use case — scanning film negatives where the densest part of the negative (brightest in the capture) rarely approaches sensor clip — the substitution does more harm than good.

## Why draft
This changes the decode for every camera RAW file. The calibration path already uses `0.0`, so there's precedent — but if the default was left at `0.75` deliberately in the main pipeline, this PR should be rejected. Opening as draft for discussion.

## Test plan
- `make all` passes (one pre-existing unrelated failure in `test_overflow_bar`)
- No logic change beyond adding the one kwarg

---
Testing the A/B toggle:

There's a "Fix WL" button in the Process panel (below Linear RAW / Narrowband / Lock). ON = fixed white level, OFF = legacy behavior.

Toggle it with a camera RAW file loaded and compare. Note: the file's brightest pixel must exceed 75% of the camera's nominal white level for there to be any visible difference — below that threshold both modes produce identical output.

---

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -1156,6 +1156,13 @@ class AppController(QObject):
         if self.state.current_file_path:
             self.load_file(self.state.current_file_path, preserve_zoom=True)
 
+    def toggle_fix_adjust_maximum(self, enabled: bool) -> None:
+        self.state.fix_adjust_maximum = enabled
+        self.render_worker.processor.fix_adjust_maximum = enabled
+        self.preview_service.fix_adjust_maximum = enabled
+        if self.state.current_file_path:
+            self.load_file(self.state.current_file_path, preserve_zoom=True)
+
     def handle_canvas_clicked(self, nx: float, ny: float) -> None:
         if self.state.active_tool == ToolMode.WB_PICK:
             self._handle_wb_pick(nx, ny)
--- a/negpy/desktop/session.py
+++ b/negpy/desktop/session.py
@@ -153,6 +153,9 @@ class AppState:
     # Transient: preview is currently peeking the flat render (not persisted).
     flat_peek: bool = False
 
+    # Temporary A/B toggle: when True, _decode_sensor_rgb uses adjust_maximum_thr=0.0.
+    fix_adjust_maximum: bool = True
+
     @property
     def local_hidden_masks(self) -> set:
         """The current file's hidden-mask indices (empty = all shown). Returns a fresh,
--- a/negpy/desktop/view/sidebar/process.py
+++ b/negpy/desktop/view/sidebar/process.py
@@ -107,6 +107,15 @@ class ProcessSidebar(BaseSidebar):
         raw_row.addWidget(self.lock_bounds_btn, 1)
         self.layout.addLayout(raw_row)
 
+        self.fix_amt_btn = self._small_toggle(
+            "mdi6.ab-testing",
+            "Fix WL",
+            self.state.fix_adjust_maximum,
+            "A/B: pin RAW white level to camera max (disable rawpy adjust_maximum_thr). "
+            "ON = fixed white level, OFF = legacy 75% threshold. Triggers re-decode.",
+        )
+        self.layout.addWidget(self.fix_amt_btn)
+
         buf_row = QHBoxLayout()
         self.analysis_buffer_slider = CompactSlider("Analysis Buffer", 0.0, 0.25, conf.analysis_buffer)
         self.analysis_region_btn = self._tool_toggle(
@@ -230,6 +239,7 @@ class ProcessSidebar(BaseSidebar):
         self.lock_bounds_btn.toggled.connect(self._on_lock_bounds_toggled)
         self.linear_raw_btn.toggled.connect(self._on_linear_raw_toggled)
         self.narrowband_scan_btn.toggled.connect(self._on_narrowband_scan_toggled)
+        self.fix_amt_btn.toggled.connect(lambda c: self.controller.toggle_fix_adjust_maximum(c))
 
         self.analysis_buffer_slider.valueChanged.connect(lambda v: self._on_buffer_changed(v, persist=False))
         self.analysis_buffer_slider.valueCommitted.connect(lambda v: self._on_buffer_changed(v, persist=True))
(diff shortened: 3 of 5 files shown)


marcinz606:
interesting, I must totally missed that param when initiall doing the raw loading. Have you noticed any difference after that change?

thetalkingdrum:
No, I barely have any camera files to test it on. I was working on a raw linear export feature when this popped up. It was Claude alerting me that there was a potential problem here. There are some comments in raw_demosaic.py so maybe @seanharding encountered it while making the sensor matrix feature?

seanharding:
I didn't knowingly run into it, but that doesn't mean it didn't impact something in that feature without my knowledge... I'm not sure I 100% grasp the issue this is correcting. I can try patching this change in later and see if it makes any difference. I've also been wanting to look at some issues I noticed with linear raw + white light scans (which I think are probably the most common). Not sure if there's any overlap between that problem and this.

thetalkingdrum:
@seanharding It seems your sensor matrix code uses the correct way of doing it (0.0), so no need to fix anything there. But Claude has left a comment in the code.

thetalkingdrum:
Here's the exact docstring text from raw_demosaic.py — the calibration-only decode module (linear_demosaic), which feeds into build_sensor_matrix in features/process/sensor.py:

"It deviates in one parameter, deliberately: adjust_maximum_thr=0.0 (see linear_demosaic). The canonical decode still runs LibRaw's default, where each frame is scaled by its own brightest pixel — harmless for a single rendered image, fatal for a meter comparing frames. Whether the canonical path wants the same fix is a separate question (it changes rendered output, so it needs its own verification); it is NOT covered here."

And the more detailed explanation, inline as a code comment right above the adjust_maximum_thr=0.0 argument itself, inside the postprocess() call:

"Scale against the camera's white level ONLY — never the frame's own brightest pixel. LibRaw's default (adjust_maximum_thr=0.75) silently switches the scaling reference to the image maximum once that exceeds 75% of the white level, so each frame is normalised by its own content. That makes the decode non-linear in exposure: rig data showed the metered base pinned at ~34400 across LED levels 128-160 (the scaling grew exactly as fast as the light), which reads as an LED/shutter defect and is neither. 0.0 disables the substitution, making the demosaiced scale a fixed multiple of the raw counts — which is what a meter measuring absolute light requires, and what CLIP_CEILING assumes."

marcinz606:
@thetalkingdrum can you put a/b button in process panel temporairly to enable/disable this change? then it's easier to test the effect

thetalkingdrum:
A/B Done! Note: the file's brightest pixel must exceed 75% of the camera's nominal white level for there to be any visible difference — below that threshold both modes produce identical output. ON = fixed white level, OFF = legacy behavior.

seanharding:
Ok, yeah, I don't see any difference with the test files I have handy. Maybe I'd need to shoot one with a higher exposure to see it...

thetalkingdrum:
Makes sense you wouldn't see it in single files. The rescale hits the whole frame, not just the bright spot, so there's nothing locally off to spot. It only shows up when you compare frames from the same roll/exposure. One frame ends up slightly different in density/brightness than its neighbors, just because it happened to have a highlight that crossed the 75% threshold and the others didn't. Invisible file by file, but exactly the kind of thing that'd make a roll feel subtly inconsistent frame to frame.

marcinz606:
I've seen some people complaining on reddit about the results comming dark, I wonder if that could be kicking in if they they were overexposing their camera scans by a lot

thetalkingdrum:
I've been thinking they just don't set the crop correctly. But could be. I would think probably the effects here are more subtle, but I have no way of testing really.

marcinz606:
> I've been thinking they just don't set the crop correctly. But could be. I would think probably the effects here are more subtle, but I have no way of testing really.

or it might be both :)

seanharding:
Batch autocrop still doesn't seem to work very well in my experience, so that may be a factor… It's better than it was, but I still end up having to tweak it almost every time.

seanharding:
(And also, as discussed in discord, I do think that people having linear raw turned on while camera scanning with white light sources will probably get too-dark results)

marcinz606:
> Batch autocrop still doesn't seem to work very well in my experience, so that may be a factor… It's better than it was, but I still end up having to tweak it almost every time.

default analysis buffer is pretty generous (5%) so even if autocrop is not exactly spot on it shouldn't catch outside white light into analysis

marcinz606:
I can't see any difference on files i have but IMO it makes sense to set this everywhere at 0.0, I've made separate PR doing just that #718 and will close this one

PR #724, opened 2026-08-01
OpenICE IR removal method

marcinz606:
Ports [openICE](https://github.com/marcinz606/openICE), a reverse-engineering of Nikon Scan's Digital ICE verified byte-exact against the original engine, as a second IR reconstruction alongside the existing one. Both are GPL-3.0, so no licensing friction.

Selected by a new **Method** combo at the top of the IR Removal panel. **Default is unchanged** — this ships as a bake-off, not a replacement.

## Why

Where the current fill averages fine structure away and pastes borrowed grain over it, ICE rebuilds the picture: working in log density, each detail band is added back only where it beats the infrared's own contrast at that same scale, so the IR channel itself decides what is photograph and what is dust.

## What had to change

The algorithm ports cleanly; its constants do not.

**The level constants are absolute and tuned to a Coolscan under Nikon Scan.** On the LS-9000 frame in `samples/dust/`, *every* pixel clears ICE's clear-film gate (τ=8847) — so its crosstalk fit would treat dust as film — and its ramp saturates at a 15% IR loss the deepest dust here (−11%) never reaches. Clear-film level, ramp and dust floor are now measured per frame from the gate's own distribution. openICE's docs concede this for flatbeds; it also bites a genuine Coolscan driven by different software.

**Added a confidence write ramp**, which ICE has no equivalent of. It rebuilds any pixel under full confidence and floors the result at the original, which half-wave-rectifies IR noise into one-sided brightening. That mottled 25% of a synthetic clean frame. With the ramp, clean film comes out bit-identical.

**Dropped the LCG dither** (the finest band already restores real grain, and it was the only serial dependency in an otherwise per-pixel algorithm) and **added a per-component route cap** — the film rebate sits above the dead floor on some scanners and arrived as one component covering 1.2% of a flatbed frame, blowing the inpaint budget and taking the real dust with it.

## Isolation

`openice.py` imports nothing from `logic.py`; the only join is one branch in `_ir_bake` plus the config field and its sidebar row. Constants that look duplicated are duplicated deliberately. Whichever method loses can be deleted as one file and one branch — noted in CLAUDE.md so it survives contact with future refactors.

## Verification

- **Against the C reference** (built from `src_C`, 64-bit): confidences and gate pyramid match to float32 noise across all four scales; give-up map agrees on 99.95%. This caught a real omission — I had dropped ICE's `θ` gate bias as negligible and it showed up as a clean 1.0000 offset, so it went back in.
- **Both `samples/dust/` scans**: OpenICE leaves clean film 100.00% untouched (current path: 99.83–99.98%) with a 4.7× smaller mean change, for comparable dust removal.
- **End-to-end through the real GUI**, headless: loads, detects IR, both methods render and differ, combo syncs.
- 21 new tests in `tests/test_ir_openice.py`; the existing 60-test IR suite untouched. `make all` green (3010 passed).

## Note for a follow-up

The existing NegPy path logs `routed defects cover 2.2% of the frame — inpaint skipped` on the flatbed sample for the same rebate reason fixed here. Left alone to keep the methods isolated, but it is a real bug in the default path.


thetalkingdrum:
Wow!

PR #752, opened 2026-08-03
Work prints: named versions of a frame that survive the undo stack

marcinz606:
## What

A **work print** is a named version of one frame's edit — the darkroom habit of keeping the prints you made on the way to the final one, so you can go back to the third attempt after deciding the fifth went too far. Master-printer workflow descriptions are consistent about this: several work prints get made and kept, and the recipe is written down.

NegPy's history is an unnamed 100-deep undo stack that prunes, so a version could not be kept, compared or re-exported by name.

In the History tab, above the edit history:

* **Save work print** (**Ctrl+Shift+S**) keeps the live edit under a name, offering *Work print 1*, *2*, … Saving over an existing name asks first.
* **Click** one to make it live; **right-click** → Export this version… / Rename… / Delete.
* The list only appears once you have saved one.

## Why a separate table, not a name column on `edit_history`

The obvious version — name a history step, exempt named steps from pruning — is a trap, and worth spelling out because it looks cheaper:

`update_config` calls `truncate_history_above` whenever you edit after stepping back. Jump to work print 1, nudge a slider, and work print 2 is deleted as an "orphaned future branch" — the name would be destroyed by the very thing it is meant to survive. Exempting named rows there instead strands them above `max_history_index`, and `history_steps` iterates `range(max_history_index + 1)`, so the panel could not show them.

A `work_prints (file_hash, name, created_at, settings_json)` table sits outside that bookkeeping entirely: nothing to exempt, no undo semantics to bend. Storage follows the `normalization_rolls` pattern in the same file, and loading goes through `update_config`, so **loading a work print is itself undoable** for free.

Per-frame, so the DB is the right home; `services/assets/presets.py` (JSON files, library-wide looks) stays separate — a preset list polluted with "frame 12 work print 2" is useless. Not mirrored to `.negpy` sidecars.

## The easy-to-miss sites

`rehome_file_settings` (so a superseded content hash carries its work prints — this is what makes `hash_migration.py` need no change), `database_stats` + the Manage Database row, and both wipe lists (`clear_saved_edits`, `reset_everything`).

## Verification

`make all` green (3256 tests). 17 new tests cover the store, the rehome, both wipes, survival of `prune_history`/`truncate_history_above`, and that loading is undoable.

End to end through the real `MainWindow` History panel on a real scan, 21 checks: the list stays hidden until there is one, both entries list newest-first, both survive 103 edits past the prune horizon *and* a truncating edit after stepping back, loading restores its density and Ctrl+Z puts back what was on screen, rename keeps the settings, the panel follows every change, `Ctrl+Shift+S` is bound with no other action claiming it, and Clear saved edits drops them.

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -26,7 +26,7 @@ The right-hand tabs are arranged in the order you actually work, which mirrors t
 | **Colour** | palette | Lab · Toning | Chroma, sharpening, effects, split/chemical toning |
 | **Finish** | brush | Retouch · Finishing | Dust removal, vignette, border, carrier |
 | **Favourites** | star | Your chosen sliders | Quick access to the controls you use most |
-| **History** | clock | Edit history | Step back through every change |
+| **History** | clock | Work prints · Edit history | Keep named versions, step back through every change |
 | **Export** | file | Export settings | Format, size, colour, batch output |
 | **Metadata** | tags | Archival metadata | Original camera/lens/film details |
 | **Scan** | camera | Scanner · Camera Scanning | Capture film directly (Linux/macOS) |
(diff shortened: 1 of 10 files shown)


marcinz606:
Fixed a bug found in review: clicking a saved work print produced empty `N · —` rows in the edit history.

Root cause was not in `load_work_print` — `session.update_config` pushes a history step unconditionally, so persisting a config identical to the live one stores a duplicate, and `history_step_label` diffs it as unchanged. Reproduced first:

```
after save          : ['0 · base', '1 · exposure']
click wp (unchanged): ['0 · base', '1 · exposure', '2 · —']
click wp again      : ['0 · base', '1 · exposure', '2 · —', '3 · —']
```

The same path is reachable without work prints (resetting an already-default panel, pasting identical settings), so the guard went into `update_config` rather than the caller. A no-op persist now also stops discarding the redo branch. A work print whose edit *does* differ still records exactly one step and stays undoable.

Covered by two new tests in `test_history_panel.py` (no step for an unchanged persist; the redo branch survives one) and one in `test_work_prints.py`. `make all` green — 3259 tests. The headless History-panel run now asserts three clicks add no rows and that no `· —` row exists anywhere in the history.

PR #756, opened 2026-08-04
Linear Output: composite support, correction toggles, and no-op fixes

thetalkingdrum:
## What

Three additions to the Linear Output export (#722):

**Composite sources** — Linear Output now handles multi-file sources:
- RGB-scan triplets merge three narrowband exposures into one TIFF
- Stitch composites assemble multi-part scans with per-part flatfield and sensor correction for clean seams
- Stitch + triplet combinations (e.g. 4-part stitch of triplets = 12 RAW files → 1 TIFF)

**Correction toggles** (camera RAW only) — three checkboxes to optionally bake corrections before writing:
- Apply white balance (as-shot gains, green-normalized)
- Apply flatfield
- Apply sensor correction
- All default to off. Hidden for Pakon/DNG. Greyed out when the correction isn't configured. Warning hint that baked corrections are irreversible.

## Testing

27 new tests (79 total): 8 triplet, 8 stitch, 11 corrections.

Verified with real Nikon Z f NEF files (single frame and stitch composite).

`docs/USER_GUIDE.md` and `docs/PIPELINE.md` updated.


thetalkingdrum:
@marcinz606 The stitch+triplet combo path (e.g. 4-part stitch of triplets = 12 files → 1 TIFF) is only covered by synthetic tests — I didn't have the triplet + stitch files raw files to test it for real.  You might have some files handy from when you built that feature, could you give it a spin?

marcinz606:
tested it, works

PR #762, opened 2026-08-05
feat: expose SANE scan-exposure-time as a slider in the Scan tab

diegotonetti99:
### Summary

When a connected SANE scanner exposes the `scan-exposure-time` option (e.g. genesys-based devices), a new **Exposure** slider appears in the Scan tab below Auto-exposure. The slider overrides the scanner's default exposure time in microseconds; the label renders the value in µs, ms or s as appropriate. A device without the option hides the slider, so a saved setting never breaks a different scanner.

### Why

Some flatbed/film scanners (notably genesys) let the operator dial in a hardware exposure time to control density on the scan. NegPy's Scan tab already had Auto-exposure, but for scanners that support a manual override there was no UI to reach it — the option sat unused in the SANE option map.

### What changed

- **Capability detection** (`negpy/infrastructure/scanners/base.py`, `sane_backend.py`): `ScannerCapabilities` gains `exposure_time_us: tuple[int, int] | None` (min/max in µs). `_detect_scan_exposure_time()` finds the option under either `scan_exposure_time` or `scan-exposure-time` key conventions, validates the constraint range, and rejects implausible values (hi < 1.2 s). `_caps_from_options()` wires it in.
- **Params & settings** (`negpy/infrastructure/scanners/params.py`, `settings.py`): `ScanParams` and `ScannerSettings` gain `exposure_time_us: int | None = None`. None = scanner default; ignored when the device has no such option.
- **Backend application** (`sane_backend.py`): `SaneBackend` applies the value via `setattr(dev, <option>, int(...))` at scan time, guarded by an existence check so a saved value targeting one scanner silently no-ops on another. Failure to set raises loudly rather than dropping the override.
- **Scan sidebar** (`negpy/desktop/view/sidebar/scan.py`): a `QSlider` + value label row, hidden unless the live device reports a usable range. Label auto-formats µs/ms/s. The value is read into `ScanParams` on every scan and persisted in `ScannerSettings`.
- **Tests** (`tests/scanners/test_capabilities.py`): `TestScanExposureTimeCapability` covers range-tuple detection, hyphenated-key detection, absence → None, wiring into caps, and the default-None path.
- **Docs** (`docs/USER_GUIDE.md`): the Scanner (SANE) entry now documents the Exposure slider and its conditional visibility.

### How to test

1. Connect a scanner that exposes `scan-exposure-time` (e.g. a genesys device).
2. Open the Scan tab — the **Exposure** slider should appear below Auto-exposure with the device's min/max range.
3. Drag it and confirm the label switches between µs/ms/s; run a scan and verify density changes with the value.
4. Disconnect and connect a scanner without the option — the slider should be hidden and a previously saved value should not raise.

### Notes

- No GPU/pipeline changes; scanner-acquisition only.
- CPU/GPU parity unaffected.

### Disclosure

Portions of this change were written with the assistance of an AI coding tool (Claude / opencode). All code was reviewed and is the author's responsibility.⏎

### DIFF
--- a/negpy/desktop/view/sidebar/scan.py
+++ b/negpy/desktop/view/sidebar/scan.py
@@ -10,6 +10,7 @@ from PyQt6.QtWidgets import (
     QLineEdit,
     QProgressBar,
     QPushButton,
+    QSlider,
     QSpinBox,
     QVBoxLayout,
     QWidget,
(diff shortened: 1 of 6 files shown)


marcinz606:
can't verify on hardware but the code makes sense, thanks!

diegotonetti99:
I have verified it on my Plustek 7500i. The new sane-devices backend exposes this new feature for controlling the exposure time, really useful for dense negatives. I had to compile the sane backend from source since the feature is not yet in the Fedora repos version.

PR #765, opened 2026-08-06
Linear Output: scanner loaders, ICE, TIFF gamma, LogLuv decode

thetalkingdrum:
## Summary

NegPy can now open and process files from every major film scanner — Coolscan NEF, Flextight FFF (including raw LogLuv-encoded files), and Noritsu RAW join Pakon and TIFF as first-class formats in the normal editing pipeline. Linear Output also gains support for all of these, plus TIFF gamma linearization and IR-based dust removal.

### Scanner loaders (new formats in the app)

- **Coolscan NEF loader** — reads SubIFD chain, rejects camera NEFs (CFA detection), no colour-space assumptions
- **Flextight FFF loader** — picks largest IFD by pixel count, parses FlexColor plist metadata (film stock, gamma, serial, scan date)
- **SGI LogLuv decoder** for raw Flextight .3fr/.fff files — LogLuv32/24 → XYZ → linear sRGB pipeline. Thanks to @rohanpandula for the decode logic in [flexcolor-tool](https://github.com/rohanpandula/flexcolor-tool) (MIT) — ported with attribution. Great timing!
- **Noritsu RAW loader** — headerless BGR16 scanner dumps with three-tier dimension detection (exact match, known-width with novel height, novel-width fallback)

### Linear Output

- **ICE dust removal toggle** — applies IR-based dust/scratch correction before writing when an IR channel is available (extra-page, ExtraSamples, or sidecar IR sources)
- **TIFF linear output** with manual input gamma selector (linear, 1.8, 2.2, sRGB) so the source encoding can be declared and linearized before export
- **IR detection fixes** — ExtraSamples tag semantics (UNSPECIFIED = IR, alpha values dropped), extra-page IR (SilverFast iSRD convention), sidecar IR with mask support
- **Output is always clean** — no ICC profiles, EXIF color space tags, or XMP color metadata from the source are copied through. Only raw pixels plus device metadata (Make, Model, DateTime)
- Fix crash on non-standard TIFF DateTime tags

### Cleanup

- **Loader cleanup** — stripped sRGB linearization assumptions from NEF and FFF loaders, removed dead IR channel logic (neither format has IR hardware), kept `_decode_tiff` independent from TiffLoader to avoid inheriting its sRGB assumptions
- Docs updated: USER_GUIDE.md and PIPELINE.md cover all new formats and controls

## Test plan

- [x] `make all` passes (3414 passed, 1 pre-existing overflow bar failure)
- [x] 13 new LogLuv unit + integration tests with synthetic LogLuv32 TIFF files
- [x] NEF/FFF/Noritsu loader tests: detection, rejection of non-scanner files, float32 output, metadata extraction
- [x] TIFF IR tests: ExtraSamples semantics, extra-page IR, sidecar IR, alpha channel drop
- [x] Manual testing: app launches and runs without errors

Scanner loaders were tested against real scan files where available (Coolscan NEF, Flextight FFF, Noritsu RAW), though the sample set was small. Only uncompressed RGB FFF files were available for testing — no real LogLuv-encoded FFF yet, so that path is validated against synthetic round-trip data only. Format handling is grounded in scanner documentation and the flexcolor-tool reverse-engineering work (which itself mirrors libtiff's reference implementation). More real-world files welcome.

**Follow-up fix (738f5e8):** Added per-channel percentile normalization to the LogLuv decoder. The original port was missing the `normalize_linear` step from flexcolor-tool — the bare `clip(0,1)` silently truncated HDR values instead of mapping them into range. The decoder now applies 0.2th/99.8th percentile normalization per channel after XYZ → linear RGB, matching the reference implementation and correcting the per-channel black/gain offset inherent in Flextight CCD data.


thetalkingdrum:
Just wanted to add that Coolscan nef, Noritsu RAW and Flextight FFF work perfectly fine in the normal pipelines as well. The title might make it seem like this was a Linear output only feature.

PR #782, opened 2026-08-09
fix(canvas): stop the overlay erasing the frame the GPU canvas painted

### LINKED ISSUE #781: black screen with GPU on since Commit 8fc35b6
getting a black screen when turning GPU on since Commit 8fc35b6. no problem on CPU. 

System is Windows 11 Pro with NVIDIA GeForce RTX 3050 (8 GB), AMD Ryzen 7 5700G, 64GB.

seanharding:
With GPU acceleration on, the canvas showed nothing at all: white on macOS, black on Windows. Linux was unaffected.

CanvasOverlay punched an alpha hole through the whole canvas rect on macOS and Windows whenever the GPU widget was visible. That is correct for a *screen* present, where a native surface sits behind the Qt widgets and the hole reveals it. rendercanvas deliberately picks a *bitmap* present on Qt, so the frame is read back and blitted into the same backing store by the canvas's own paintEvent; the overlay paints on top afterwards and wipes it. What remains is the bare window background, which is light on macOS and black on Windows.

The hole is now punched only when there really is a native surface underneath, which GPUCanvasWidget.presents_to_screen() reports from the canvas itself rather than from a platform guess.

The compositing has been wrong since it was written; it only became visible when the soft proof moved into the display LUT. Before that, a proofed render was baked to a host array on the render thread, and soft proofing is on by default, so every default frame took the CPU/QPainter branch of ImageCanvas.update_buffer and the GPU widget was never in the path.

Verified against the real app: compositing the canvas the way the screen does gives a frame that matches the presented bitmap (mean 95.2 / std 98.0 against 1.4 / 11.0 before), and zoom and pan leave no residue — the canvas re-blits under every overlay repaint, which is what clears it now that the fill is gone.

Fixes #781

### DIFF
--- a/negpy/desktop/view/canvas/gpu_widget.py
+++ b/negpy/desktop/view/canvas/gpu_widget.py
@@ -184,6 +184,17 @@ class GPUCanvasWidget(QWidget):
             address_mode_w=wgpu.AddressMode.clamp_to_edge,
         )
 
+    def presents_to_screen(self) -> bool:
+        """True when the frame goes to a native surface of its own, which the overlay
+        must punch an alpha hole to reveal.
+
+        rendercanvas picks a *bitmap* present on Qt: the frame is blitted into the
+        shared backing store by the widget's own paintEvent, and a hole punched over
+        it erases the frame instead of revealing anything.
+        """
+        sub = getattr(self.canvas, "_subwidget", self.canvas)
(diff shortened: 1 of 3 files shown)


dtrtuser:
thanks, works like a charm now. and great performance too!

PR #784, opened 2026-08-09
fix(preview): give the navigate-back memo its pixels back

marcinz606:
Switching to a frame viewed a moment ago blanked the canvas and ran the "Processing" spinner, where it used to repaint instantly. Regression from #777.

## Cause

#777 moved the soft proof into the display LUT, so a render publishes a `GPUTexture` instead of a host array on the GPU path — the primary one. Two consumers depended on it being an array:

- The memo's only store site (`controller.py`) is guarded by `isinstance(result, np.ndarray)`, so it stopped storing entirely and every navigate-back missed → spinner on every switch. It still worked with the GPU disabled.
- `load_file` emits `gpu_textures_released` so the canvas stops sampling a pooled texture the engine is about to free. With a host array the previous frame used to stay up dimmed under the spinner — which is what `main_window._on_loading_started` still claims to do. With a texture there was nothing to fall back on, so the canvas went black.

Neither the memo's own code nor its store site changed in #777; the premise under them did.

## Fix

The memo now holds the texture itself. A file switch destroys the whole pool anyway, so the displayed texture is handed over instead of destroyed: `GPUEngine.cleanup(retain=tex)` skips it and drops its pool key, and the next render allocates a fresh one. No readback (~10 ms preview / ~695 ms HQ), no copy. `RenderMemo` frees what it owns on eviction, overwrite, `invalidate` and `clear`.

A GPU result cannot be filed at render time — the pool overwrites that texture on the next frame, and `select_file` has already hydrated the *incoming* file's config by the time `load_file` runs, so the outgoing memo key is unrecoverable there. Only the identity is kept; `load_file` files the texture under it on the way out.

Retention is refused while a render is in flight or queued: it paints into that same pooled texture, so the pixels would stop matching the key they are filed under. Falling back there costs today's spinner, not correctness.

With the texture retained the canvas keeps sampling it, so the previous frame stays up under the dimmed spinner while a genuinely new frame loads.

## Notes

- Memo VRAM is capped by `preview_cache_max_full_res_entries` (default 2 = current + previous), the same knob as the preview cache. The pre-#777 memo held two host arrays of the same size.
- The HQ toggle now clears the memo: resolution is part of the key, so every entry would be a permanent miss holding a full-size texture.
- Memo-owned textures are freed before the device is destroyed at shutdown.

## Tests

- `tests/test_navigate_back_memo.py` (new) — a settle GPU render is filed on the way out; compare/splash/tool/drag frames are not; a late render of the outgoing file is not; the CPU path still files its array directly; retention refused mid-render and while queued; the identity is spent once.
- `tests/test_render_memo.py` — owned textures freed on evict/overwrite/invalidate/clear; re-storing a frame served *from* the memo does not free its own texture; `rekey` keeps it alive; array payloads untouched (the test-strip memo shares the class).
- `tests/test_gpu_engine.py` — `cleanup(retain=tex)` leaves it alive, drops its pool key, and the next render gets a different texture.

`make all` green (3627 passed, 4 skipped).

### DIFF
--- a/negpy/desktop/controller.py
+++ b/negpy/desktop/controller.py
@@ -276,7 +276,7 @@ class AppController(QObject):
     flat_peek_changed = pyqtSignal(bool)
     zoom_requested = pyqtSignal(float)
     zoom_changed = pyqtSignal(float)
-    _render_cleanup_requested = pyqtSignal()
+    _render_cleanup_requested = pyqtSignal(object)  # texture to spare, or None
     status_message_requested = pyqtSignal(str, int)
     status_progress_requested = pyqtSignal(int, int)
     batch_started = pyqtSignal(str, bool)  # title, abortable
@@ -429,6 +429,9 @@ class AppController(QObject):
         # Last displayed render per frame — navigate-back paints it instantly
         # while the authoritative render refreshes underneath.
         self._render_memo = RenderMemo()
+        # (source_hash, memo_key, content_rect) of the settle render whose GPU texture
+        # is on screen; load_file files it under this identity on the way out.
+        self._last_render_identity: Optional[tuple] = None
         # Test strips, keyed density/grade-blind (see _strip_memo_key).
         self._strip_memo = RenderMemo()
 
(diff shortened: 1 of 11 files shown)


marcinz606:
## Verified against the real app

Headless driver (`NEGPY_USER_DIR` sandbox, `QT_QPA_PLATFORM=offscreen`, real GPU), stepping A→B→A→B→A through `session.select_file`, counting `loading_started` and `gpu_textures_released` per leg:

| leg | spinners | canvas releases | buffer |
|---|---|---|---|
| A cold | 1 | 1 | GPUTexture |
| B cold | 1 | 0 | GPUTexture |
| A back | 0 | 0 | GPUTexture |
| B back | 0 | 0 | GPUTexture |
| A again | 0 | 0 | GPUTexture |

Every navigate-back is a memo hit: no spinner, and the canvas is never told to let go, so no black flash. No `Texture has been destroyed`, no wgpu validation error, no traceback across the run.

Same walk with the GPU disabled: memo hits as before (0 spinners on every return leg), `base_positive` an ndarray. `gpu_textures_released` still fires there and is harmless — it clears the hidden GPU widget while the overlay keeps its host image.

### One thing the run turned up, pre-existing and left alone

`_render_memo_key` hashes the whole `config.to_dict()`, export fields included — `export_path`, `userDir`, `export_color_space`, `overwrite`, `export_resolution_mode`. None of them change a rendered pixel. It only shows up when a `.negpy` sidecar carries export paths from another machine: the first render uses the sidecar's values, the DB copy normalises them, and the first navigate-back to that frame misses. Every later one hits. Narrowing the key to the fields that shape pixels is a separate change.

marcinz606:
## Follow-up: the memo now holds eight frames, not two

It shared `preview_cache_max_full_res_entries` with the full-resolution decode cache, so it kept the current frame and the one navigated from — nothing further back. Stepping three frames along a roll and returning re-rendered.

Preview-size frames are ~27 MB of VRAM each at `preview_render_size=1600`, so eight cost ~218 MB. Two things stay on the old conservative budget: HQ renders (hundreds of MB apiece) and test-strip entries (four mosaics apiece).

New knob, settable in `override.toml` under `[performance]`:

```toml
# render_memo_max_entries = 8
```

Re-verified with the same headless walk: every return leg still 0 spinners, 0 canvas releases, no validation errors. Full suite green (3630 passed).

PR #785, opened 2026-08-09
feat(update): install a new release from inside the app

Aa0rR:
The "Update Available" notice sent you to the releases page to download, uninstall and reinstall by hand. It now opens an update window — release notes, download size, one button — that fetches the build matching how this copy was installed, closes NegPy, installs over the old version and reopens on the new one.

Per platform: the NSIS installer runs silently over the current install (elevated, since it lives in Program Files, and restarted through Explorer so the app itself does not inherit the admin token); the macOS DMG is mounted and its bundle swapped in where the running one sits; the Linux AppImage file is replaced and relaunched.

A running program cannot overwrite itself, so the swap is a detached script that starts only once this process is gone. It waits on the pipe NegPy holds rather than polling the PID — an exited-but-unreaped process is a zombie, and a zombie still answers `kill -0`. Nothing is replaced before then, so a failed download or a refused UAC prompt leaves the install as it was, and a directory the user cannot write is reported instead of half-applied.

Where NegPy cannot swap itself — a source checkout, a release with no build for the platform — the notice still appears and the button opens the releases page.

Also:
- installer.nsi records its install directory, so a silent upgrade lands on top of an existing install wherever the user put it.
- The check thread belongs to the module, not to the panel that starts it: a QThread destroyed while a stalled socket read is in flight aborts the process.
- New unbound "Check for updates" action in the shortcut registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ It is built with **Python**, running natively on Linux, macOS, and Windows.
 ### Download
 Grab the latest release for your OS from the **[Releases Page](https://github.com/marcinz606/NegPy/releases)**.
 
+After that NegPy keeps itself current: when a new release is out, the left panel shows an **⬇ Update Available** link that downloads and installs it for you, then reopens on the new version. No manual download, uninstall or reinstall.
+
 #### Linux
 I provide an `.AppImage`. Make it executable using `chmod +x` and It should just work.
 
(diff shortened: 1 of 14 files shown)


thetalkingdrum:
Finally got to try this out and it worked perfectly on macos. Didn't have to do any xattr -cr /Applications/NegPy.app shenanigans for the update to work :)

PR #796, opened 2026-08-10
fix: apply DNG calibration tags in Linear Output's 3-channel decode

thetalkingdrum:
## Summary

#786 taught `RawpyLoader`'s import path to correctly decode 3-channel LinearRaw
DNGs by replaying their `LinearizationTable`/`BlackLevel`/`WhiteLevel` tags
(`_peek_linear_dng_rgb`), fixing import for DNG 1.7 JPEG-XL files from DxO
PhotoLab/PureRAW and Lightroom Enhance. It explicitly flagged one thing it
didn't fix:

> **Known follow-up, not fixed here:** `services/export/linear_output.py`'s
> Linear-export path for 3-channel LinearRaw DNGs reads the same SubIFD
> without applying `LinearizationTable`/`BlackLevel`/`WhiteLevel`, so a Linear
> export of this file class comes out ~20x too dark.

This closes that gap. `_decode_dng`'s 3-channel branch now calls
`_peek_linear_dng_rgb` directly instead of re-deriving its own (incomplete)
copy of the same SubIFD read — one tag-aware decode shared by both the import
and export paths instead of two.

Note: the bug (and the fix) apply to any 3-channel LinearRaw DNG with real
`BlackLevel`/`WhiteLevel`, not just JPEG-XL-compressed ones. `_decode_dng`'s
3-channel branch has never gone through libraw — before this change or after,
it's always been a direct `tifffile` read — so it was never actually gated on
compression. JPEG-XL is just how #786 happened to surface it (it's the
compression libraw can't open at all, which is what put the fallback path
under scrutiny). An uncompressed 3-channel LinearRaw DNG from any other
camera-oriented DNG writer that declares real `BlackLevel`/`WhiteLevel` had
the same ~20x bug before this fix and is equally fixed now. A genuine
Bayer/CFA camera DNG (not LinearRaw at all) is unaffected either way — it's
routed to `_decode_camera_raw`, the same path used for CR2/ARW/NEF, and never
reaches this code.

Two smaller things fell out of reusing the shared helper instead of the
inline read:

- **Error type.** The peek function's post-tag-read math (linearization
  lookup, black/white normalization, crop) isn't covered by its own internal
  try/except, so a malformed tag combination could previously propagate an
  unwrapped exception instead of the `ValueError` the old inline code always
  raised on failure. Wrapped the call so that contract holds either way.
- **IR/RGB alignment.** `_peek_linear_dng_rgb` applies `DefaultCropOrigin`/
  `DefaultCropSize` when present, but the SilverFast HDRi IR-plane lookup
  (`_peek_hdri_ir_page`) matches against the *uncropped* sensor dimensions.
  A file with both a populated crop and a same-file IR page would previously
  export a cropped RGB TIFF paired with a full-sensor `_ir.tif` — spatially
  misaligned. Added a shape guard that raises instead.

## Scope

Export only — `services/export/linear_output.py`. No changes to import, the
4-channel VueScan/IR path, or the Expansion control.

## Investigation

Before touching the code, checked whether SilverFast/VueScan scanner DNGs
populate these same calibration tags — if they did, the existing manual
**Expansion** override might be redundant. They don't: three real SilverFast
HDRi samples (Kodak Vision3 250D, Plustek OpticFilm 8100, SilverFast 8.8.0)
carry none of `BlackLevel`/`WhiteLevel`/`LinearizationTable`/`DefaultCrop*`.
Per the DNG spec, absent tags default to "assume full container range" —
exactly what the old naive code already did — so this fix is a verified
no-op for that writer, and Expansion stays exactly as-is. A fourth real file
from a different tool ("openICE (open reimplementation)", per its `Software`
tag) showed the same pattern.

## Testing

- `tests/test_linear_output.py::TestDng3ChTagAwareDecode` (8 tests):
  `BlackLevel`/`WhiteLevel` correctness (uncompressed fixture — this is the
  test that proves the ~20x bug itself, independent of JPEG-XL), no-tags
  no-op, `DefaultCrop*` cropping, crop+mismatched-IR-page raising, a
  non-trivial `LinearizationTable`, the exception-wrapping fix, the
  not-found case, and an end-to-end run against a genuinely
  JPEG-XL-compressed (`compression=52546`), thumbnail+SubIFD-shaped fixture
  matching a real DxO/Lightroom Enhance file specifically — the same fixture
  shape `tests/test_raw_handlers.py` already uses to verify
  `_peek_linear_dng_rgb` and the import fallback directly, run here through
  `export_linear_output` instead.
- Verified against four real 3-channel LinearRaw DNGs from two different
  writers (three SilverFast HDRi, one openICE) that output is unchanged to
  within 1 LSB out of 65535 — float64 vs. float32 intermediate rounding in
  the shared decode, not a behavior change.
- No real DxO PhotoLab/PureRAW or Lightroom Enhance sample was available to
  test the black/white-level correction against directly; the JPEG-XL
  fixture above is the closest synthetic stand-in for that specific writer.
  `_peek_linear_dng_rgb` itself (the function this PR now reuses, unchanged)
  was verified against a real DxO PhotoLab 9.6 sample in #786.
- `make all` clean, aside from the same pre-existing, unrelated failure in
  `test_overflow_bar.py` #786 also noted (confirmed failing identically on
  unmodified `main`).


marcinz606:
nice with issue templates!

thetalkingdrum:
Haha, yes I was thinking of how to minimize all the issue reports with
zero of the info needed to reproduce:)

That was just a quick mock up, but I think the idea holds some merit in
some shape or form.

man. 10. aug. 2026 kl. 19:23 skrev marcinz606 ***@***.***>:

thetalkingdrum:
It somehow got included here by accident, but maybe that was a happy
accident :)

man. 10. aug. 2026 kl. 19:36 skrev Mats Fagerberg ***@***.***>:

PR #797, opened 2026-08-10
feat(retouch): trace a transport scratch from one click

### LINKED ISSUE #788: [Feature request] Retouch for film scratches from left to right
It would be great to have a retouch tool for film scratches. These typically run along the direction of film transport, so they appear as straight lines crossing the entire frame. Often across every frame on the roll, since they're caused by the camera or lab transport path. 
A dedicated tool could work much faster/better than using the brush tool. The brush tool works very inconsistent on such cases.

marcinz606:
Closes #788.

## The report

Film picks up long straight scratches running through a camera or lab transport path — they cross the whole frame, often in the same place on every shot of a roll. The brush "works very inconsistent on such cases".

## Why the brush fails — measured

Spread along its length, a transport scratch is far too faint to separate from grain at any single point:

| scratch depth | per pixel | integrated along the line |
|---|---|---|
| 30% | 3.2σ | 44.5σ |
| 10% | 1.3σ | 15.8σ |
| 4% | 0.7σ | 6.1σ |

The manual-heal seed bar is 8σ. A 10% scratch reads **1.3σ per pixel** — no per-pixel gate can find it, at any threshold, which is precisely why these need a tool of their own rather than a better brush. Gain from integrating along the line: **×6–14**.

## The tool

**Click once on the scratch.** Everything else is measured:

- **Cross-section band-pass** (grain is finer, image structure broader), normalized by a **local** noise scale — a global one lets one busy corner of the frame set the bar and buries a faint scratch running through smooth sky.
- **The slope is fitted, not assumed.** Film is rarely square to the sensor; over a 5000 px frame even 0.4° drifts 35 px, which would smear the ridge across dozens of rows and destroy the very integration this depends on. On the sample the real scratches measure 0.03–0.44°. The fit is pulled toward the click, or it snaps to whatever ridge is strongest nearby instead of the one under the cursor.
- **Extent is measured.** These fade in and out, so only stretches where the ridge persists are repaired. A click that finds nothing reports it rather than touching the frame.

The output is a mask at the scratch's own width, which takes the same skirt pad and the same shared fill as every other defect. **The repair path from #795 is reused whole** — this adds a detector, not a second repair.

## Deliberately not included

**Auto-detection.** On the sample frame the strongest full-length ridge (241σ) was a *horizon* — 13 px wide and smooth, not a scratch. Separating content from scratches reliably needs cross-frame recurrence ("often across every frame on the roll"), which a per-frame tool cannot see. A click sidesteps the whole problem.

## Verification

- `make all` green; +350/−21 lines.
- New tests: the tracer finds a faint sloped line and recovers its slope to <0.002; clean film yields no line; a traced line repairs the scratch while leaving film a few rows away byte-identical; a half-clean frame is scored only where the scratch is; token + serialization + preset-exclusion.
- On `samples/scratch/202500813_072.dng` (thanks for the sample): three clicks remove three real scratches, grain intact, no residue.
- Driving the real app headlessly: click → trace → repair, confined to a 20-row band across the frame.

## Note

Extent is re-measured at each render resolution, so a preview and a full-res export can repair slightly different stretches of the same line — the same property the IR path already has, since detection follows the buffer it repairs.


marcinz606:
Follow-ups pushed (`4ad5740`).

**Wider scratches.** The band was a fixed width, so anything thicker than a hairline kept its flanks. It is now grown outward from the line by the same hysteresis the brush uses — while the ridge holds, stopping where it breaks:

| scratch | fixed band | grown band |
|---|---|---|
| 1–2 px | 16 % residual | 16 % |
| 4 px | 44 % | 17 % |
| 6 px | 77 % | 18 % |
| 8 px | 94 % | 22 % |
| 12 px | 100 % | 16 % |

Growing rather than measuring a width in pixels is also what keeps a stored line valid at any render resolution — a width read off the ridge profile disagrees with itself between preview and export, because the band-pass kernel's own width is a constant number of *pixels*, not of film. I built that measurement first and threw it away for exactly that reason.

A second thing was blocking wide ones: a full-width line a few px tall already exceeds the inpaint routing budget, so the structure-following repair was refused precisely where it was needed. That budget guards an *automatic* detector against a misregistered IR plane calling half the frame a defect; hand-placed repairs now bypass it. 16 px+ still degrades — that is a gouge, past this tool's reach.

**A guide.** Hovering traces the scratch under the cursor and draws the line plus the band it would repair, before committing. Committed lines draw the same way instead of as hairlines. Tracing is a full-frame slope search, so it runs on a debounce.

**Line Sensitivity**, in the slot the brush size occupies. Sets the ridge bar for both extent along the line and the band grown from it, so it trades reach against picking up film either side. It re-measures lines already placed, so a scratch can be traced first and tuned after. Default sits mid-range and reproduces the previous behaviour exactly.

`make all` green. New tests cover the sensitivity trade, the bar's monotonicity, wide-scratch repair, the budget bypass, and the hover guide.

PR #800, opened 2026-08-11
Attribute a finished render's thumbnail to its own frame

seanharding:
## The bug

The filmstrip sometimes shows one frame's picture on another frame's thumbnail, and it stays wrong until that cell is clicked and re-rendered.

Renders are asynchronous and a full-res decode takes seconds, so one can land after the user has clicked on to the next frame. `_update_thumbnail_from_state` took the pixels from `last_metrics` — the render that just finished — but keyed them by `uploaded_files[selected_file_idx]`, whichever frame is selected *now*:

```python
file_hash=asset_thumbnail_key(self.state.uploaded_files[idx])   # the selection, now
buffer  = metrics["base_positive"]                              # the render that just finished
```

Select A, start its render, click B before the decode finishes, and A's pixels get filed under B's key.

That this can happen is already known a few lines below, where the render memo guards for exactly it:

```python
memoizable = bool(metrics.get("memo_key")) and metrics.get("source_hash") == self.state.current_file_hash
```

The memo path had the guard; the thumbnail path never got one.

## The fix

The thumbnail now follows the render's own `source_hash`. Looking the asset *up* by that hash, rather than merely skipping the update when it mismatches, means a late render still refreshes the correct cell instead of being dropped.

With no `source_hash` it falls back to the selection. That is the `active_file_changing` caller, which fires *before* `selected_file_idx` moves — the outgoing frame is still selected there, so that path was never affected and keeps working.

## Testing

`make all` clean. New `tests/test_thumbnail_attribution.py` covers the race (a render for A landing while B is selected updates A), the ordinary case, the no-hash fallback, and a render whose frame has been unloaded. Two of them fail against the previous code.

`tests/test_thumbnail_readback_thread.py`'s stub gained the new helper; no behaviour change there.


thetalkingdrum:
Great! This has been bugging me for some time.

PR #809, opened 2026-08-11
Decode a slide neutral, and a bracket on one white balance

seanharding:
**Summary:** merged slide brackets printed colour rings around blown highlights. The E-6 as-captured decode was obeying the Linear RAW toggle the Process panel documents as inert, so frames of one bracket decoded on different white balances and the solved exposure ratios absorbed the difference — 0.75 EV instead of 1.00 on the frame carrying the highlights. Every site that decides this now asks one helper. Also fixes two latent bugs found on rebase (Batch Analysis measuring in the wrong white balance, and slide prefetch missing its cache key), and limits the merge action to transparencies.

The Process panel documents Linear RAW as inert on the E-6 as-captured path -- the camera matrix folds the as-shot multipliers back in itself -- but the decode read the stored flag regardless. So a hidden, sticky toggle decided whether white balance was applied, which is the trap the panel's own note says it was made inert to avoid.

It is a bracket that pays for it. `use_camera_wb` applies each *file's* own multipliers, and a camera left on auto white balance records different ones per frame: on an 8-frame slide bracket the darkest frame recorded B/G 1.41 against ~1.8-2.1 for the rest. The frames then sit on different scales, and `pair_ratio` cannot tell an exposure difference from a white-balance one, so the solved ratio absorbs both. That bracket's shortest link came out 0.75 EV instead of 1.00 -- and the shortest frame is the only one still carrying detail beside a blown sun, so the error landed exactly where it shows. It printed as contour rings around the sun, which is a per-channel gamut crossing: the mis-scaled frame pushes R past white over a wider radius than G, and G than B.

`effective_linear_raw` is now the one question every site asks: the decode, the CPU matrix, the GPU matrix, the source token and both preview paths. The decode and the matrix have to agree -- apply white balance at one and not the other and the render is tinted by the raw green-to-red ratio, roughly 2:1 -- and there were seven places deciding it independently.

A bracket's frames additionally decode on the reference frame's white balance rather than their own. Redundant while the transfer path decodes neutral, and deliberately so: a bracket on mismatched white balances solves wrong ratios and reports nothing.

The same bracket now solves to within 0.05 EV on every frame, and the rings are gone. Existing E-6 edits shift slightly -- under 1/255 across most of a frame, more where a channel was already clipped -- and the render still matches libraw's own conversion to within 0.25%, against the 0.1% #799 was validated at.

Merging is also now offered on transparencies only: hidden on C-41, shown disabled on B&W. A colour negative holds 5-6 stops between base and Dmax and an ordinary B&W negative nearer 4, both inside one capture; only a transparency's 10-12 needs a bracket. B&W is disabled rather than hidden because reversal monochrome (Scala, dr5, Fomapan R) does have the range and is merely not wired up.

Left alone: `sensor_unmix_available` still gates on the stored flag, so the unmix stays unavailable on the transfer path even though that decode is now neutral enough to carry it. Enabling it there is a separate decision about a correction, not a decode.


seanharding:
Looks like this needs some fixes after other recent changes. Taking a look now.

seanharding:
Alright, I think it should be good to go as long as nothing else changes in the mean time.

Issue #434, opened 2026-07-09
Crash on MacOS


InvisibleFunction:
Dragging a tiff to the sidebar results in a crash:

```
Fatal Python error: Segmentation fault

Thread 0x000000016e867000 (most recent call first):
  File "threading.py", line 359 in wait
  File "threading.py", line 660 in wait
  File "wgpu/backends/wgpu_native/_poller.py", line 101 in run
  File "threading.py", line 1044 in _bootstrap_inner
  File "threading.py", line 1015 in _bootstrap

Current thread 0x00000001f9d7de80 (most recent call first):
  File "negpy/desktop/session.py", line 227 in refresh
  File "negpy/desktop/controller.py", line 329 in _on_thumbnails_finished
  File "negpy/desktop/main.py", line 126 in main
  File "desktop.py", line 29 in <module>

Extension modules: PyQt6.QtCore, PyQt6.QtGui, PyQt6.QtWidgets, numpy._core._multiarray_umath, numpy.linalg._umath_linalg, PIL._imaging, PIL._imagingcms, rawpy._rawpy, numba.core.typeconv._typeconv, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._pcg64, numpy.random._generator, numpy.random._mt19937, numpy.random._philox, numpy.random._sfc64, numpy.random.mtrand, numba._helperlib, numba._dynfunc, numba._dispatcher, numba.core.runtime._nrt_python, numba.np.ufunc._internal, numba.experimental.jitclass._box, markupsafe._speedups, _cffi_backend, PyQt6.QtOpenGL, PyQt6.QtOpenGLWidgets, PIL._imagingmath (total: 29)
Segmentation fault: 11     /Applications/NegPy.app/Contents/MacOS/NegPy
```

This is on MacOS 26.5 (25F71) and NegPy v0.35.0

marcinz606:
I think it might be macos-specific regression from multithreading changes in 0.35.0

marcinz606:
can you checkout `fix/434-gpu-display-texture-uaf` branch to verify if it fixes it? I don't currently have macbook on hand, only linux and there there is no issue with threading

seanharding:
[...]

PR #810, opened 2026-08-11
Make multi-core CPU rendering a setting people can try

seanharding:
`cpu_parallel` already existed under [performance] in override.toml. It now also has a menu entry beside GPU Acceleration -- the same question, which processor does the work -- and applies at once, since every kernel is compiled serial and parallel and dispatched per call.

Resolution is override.toml, then the saved setting, then the platform default, with None falling through at each level, so an installation that never touches the control keeps exactly what its platform does today: Windows and Linux on, macOS off. Surfacing a control must not silently re-decide it for anyone, and a test asserts that per platform.

**The marker.** Numba's workqueue layer terminates the process on concurrent entry -- no exception to catch, no dialog, nothing in the log -- so the app cannot report that crash as it happens, only notice afterwards that the previous session never finished. It now records a clean-exit flag alongside whether parallel kernels were live, and offers to turn the setting off after a session that did not exit cleanly. Without it the setting is un-diagnosable: someone enables it, hits an abort weeks later, and reports a crash nobody can connect to it.

**The guard.** The lock holds only while *every* parallel kernel is dispatched through it; one compiled `@njit(parallel=True)` directly would bypass it and reintroduce the abort as a rare, unattributable crash. Nothing does that today, and a test now walks the package for `parallel=True` outside parallel_njit, for prange in a file that does not use it, and for direct calls to the parallel variant. Checked against a planted violation: it fails by file and line. That is the one failure mode a wider beta would surface least reliably and static analysis settles completely.

macOS stays off pending wider evidence. The lock was verified there -- the unguarded path aborts within a second, the guarded one survived 68,720 concurrent invocations across 8 threads -- but that is one machine.

What the setting is worth is stated honestly, because the kernel figures oversell it: the kernels run 5-8x faster at every array size tested, but a merge is dominated by RAW decoding, which this does not touch. On a 6-frame 24 MP bracket that is ~6.6s of a ~9s merge, so the whole operation comes down by about 10%. Three explanations for the gap were measured and rejected first: preview-scale arrays being too small (parallel still wins 5.3x), the gate serialising threads (still wins 5.6x at every thread count), and the thread pool starving libraw (decode unchanged at 6.66s against 6.54s).


thetalkingdrum:
I've had two hard crashes on my M1 MBP in the last few days. Hard as in have to power cycle hard. I wonder if this plays a role or if it's GPU related.

seanharding:
Hmm. I'd be surprised if this could cause an actual OS-level crash, but I suppose anything is possible. Any obvious pattern to what you were doing when it happened?

seanharding:
FWIW I'm also on an M1-class (M1 Pro) MBP and I haven't seen any app crashes, let alone OS crashes. But maybe I'm just not hitting the right code path to trigger it.

thetalkingdrum:
The last crash was loading a trichrome just after auto-updating to 0.51. I agree that it seems unlikely to cause an OS-level crash. GPU feels intuitively more like kernel panic territory.

thetalkingdrum:
M1 Pro here as well.

thetalkingdrum:
Did some more digging. It seems to be the same as #434

PR #814, opened 2026-08-12
Keep the import progress popup quiet while Hot Folder polls

MarinChen99097:
## Problem

Hot Folder polls the current folder every 2 s. Each frame a tethering app (NX Tether etc.) or scanner drops in starts a per-frame **"Hashing files"** and then **"Generating thumbnails"** batch, and the batch progress popup `show()`s and `raise_()`s itself on every one of them. During a tethered scanning session the "Working…" dialog pops over the canvas on every shutter release — the same "reads as *wait here* mid-roll" problem the Scanlight live view already works around locally in `scanlight.py` (`_keep_scan_windows_on_top`).

## Fix

Gate the popup at the `batch_started` connection in `MainWindow`: while the Hot Folder toggle is on, the quick per-frame import batches — the only non-abortable ones — do not show the popup. The HUD status line over the canvas still reports them ("SCANNING FOR ASSETS…", "GENERATING THUMBNAILS…"), so the import is never silent.

Every abortable batch (export, autocrop, normalization, stitch, HDR merge, contact sheet) shows the popup exactly as before, and with Hot Folder off nothing changes.

Also documents the behaviour in the Hot Folder bullet of `docs/USER_GUIDE.md`.

## Testing

- `ruff format` / `ruff check`: clean
- `ty check` (Makefile flags): clean
- `pytest tests/test_batch_progress.py tests/test_dock_panels.py tests/test_empty_state_and_header.py`: 19 passed
- Full suite on Windows 11: 3903 passed — the 11 failures present are identical on an unmodified checkout of `main` on the same machine (platform-specific: `test_parallel_gate`, `test_updater` sh scripts, ICC, etc.), so this change introduces none.

Context: found while tethered-scanning on Windows with NX Tether dropping NEFs into a hot folder — the popup raised on every shutter release.


marcinz606:
merged, thanks for contribution!

PR #815, opened 2026-08-12
Say "1 file", not "1 file(s)"

seanharding:
**Summary:** fifteen user-facing messages used the "(s)" shortcut, so a single item read as unfinished — "Reopen your last session (1 file(s))?". They now agree with their count, verbs included, via one small helper.

`count_of(n, "file")` gives "1 file" / "3 files"; `plural(n, "is", "are")` covers the verb, and an irregular plural can be passed for nouns an -s does not fit. Zero takes the plural, as English does.

Fixed across the session-restore prompt, the batch export and contact sheet confirmations, Auto Crop All, Linear Output, sidecar writing, scan batch progress, the strip preview's clamp/cut/failure notices, the sensor calibration warning and the filmstrip's Keep/Reject entries. Two places that already pluralised by hand — the export-preset confirmation's preset_word/file_word — now use the helper instead, since a second way of doing it is how the first one came to be missed elsewhere.

A test walks the package for "(s)" in any string literal that is not a docstring. It parses rather than greps: a regex over source lines matches across two separate literals — `"sizes": [list(s) for s in f["` reads as one string to it — and cannot tell a docstring from a label, which produced four false positives on the first attempt.


marcinz606:
I never noticed that :)

PR #823, opened 2026-08-14
Improve Auto Crop and Batch Autocrop

### LINKED ISSUE #828: [Bug]: Auto Crop leaves most of the border on non-35mm and wide-border scans
Auto Crop works well on 35mm camera scans shot to fill the sensor, which is what it has mostly been tested against. On a sample set of scans from other people's cameras, scanners and holders it does poorly, in two distinct ways.

### Square and medium-format frames are cropped to 3:2

The Ratio control defaults to `3:2` and Auto Crop obeys it, so any frame that is not 35mm gets center-cropped to a strip. On the sample set:

| frame | format | picture kept |
|---|---|---|
| 6x6 negative | 1:1 | 34% |
| 6x6 negative | 1:1 | 36% |
| 645 negative | 5:4 | 59% |

The 6x6 cases lose a third of the frame's width, and nothing about the result suggests the ratio setting is the cause.

### Frames with a wide border are barely cropped at all

Where the film sits well inside the capture — a holder leaving a wide margin, a sprocket rail in shot, or the neighbouring frame caught in the same exposure — Auto Crop leaves most of the border in place:

| frame | result |
|---|---|
| 35mm slide, sprocket rails both sides, adjacent frames top and bottom | 98.9% of the capture kept, i.e. effectively uncropped |
| 6x6 negative, wide grey rebate band | full rebate band left along the top and right edges |
| 35mm negative, ~8% of the width in film base and holder mask on one side | roughly a third of that border removed |
| sprocket-exposed 35mm | sprocket holes left along the top and bottom |

Common thread: these borders run two to three times deeper than a frame shot tight to the sensor, so whatever depth the border detection is currently willing to reach is not enough for them.

### Notes

Both are visible with Auto Crop on a single frame; Batch Autocrop inherits the same behaviour. The affected files span Nikon, Fuji and Olympus raws, so this is not decoder-specific — it is the shape and thickness of the border that these scans have and the tested rolls did not.

seanharding:
An attempt at general improvements to auto crop and batch auto crop.

I'm opening this as a draft because I can only test the scans I have from my setup. This overall creates much better (if still not perfect) auto-crops for my examples, but it needs to be tested against more scans.

  ## Why it was failing

  A frame shot to fill the sensor compresses the holder mask and the film-base sliver into                                                                                                                                            
  a few pixels. The existing refinement probes a ring 4% of the film box across, so it
  straddled both tiers plus the picture and abstained — on 27 of 36 frames of the first
  roll. Auto Crop then returned essentially the whole frame.

  The deeper problem was that borders were looked for at *fixed* brightness levels. On a
  thin negative the frame's own highlights set the exposure of the scan, which leaves the
  rebate under any fixed bright level and a holder edge above any fixed dark one. Both are
  invisible to an absolute threshold and obvious relative to the picture beside them.

  ## What changed

  - **The border is walked line by line** (`measure_film_edges`), which tolerates the
    slight tilt that puts mask, base and picture on the same row. Four kinds of border are
    recognised: mask and bare bed by level; a rebate brighter than the picture and a holder
    edge darker than it, measured against the picture on that line. The walk bridges the
    ramp between tiers, so a crooked holder with its mask *outside* the rebate doesn't hide
    it.
  - **A blown highlight no longer passes as the light bed.** The film-box sanity check
    asked only that the surround be bright; a thin negative's picture clears that. Two
    frames were being cropped to 25% and 46% of their area.
  - **Batch Autocrop takes the frame's own evidence** where it has it: the rebate width is
    pooled per side rather than discarded wholesale, a frame with no film box still measures
    its edges, a frame isn't dropped because a busy picture out-shouts its film edge, and
    deskew isn't pooled at all — tilt is not a roll property, and the line fit was being
    double-counted against the contour angle.

  ## Results

  Median frame kept, five rolls:

  | roll | single-frame | Batch |
  |---|---|---|
  | 2026-08-12-01 (36) | 94.1% | 94.7% — 36/36 resolved |
  | 2026-06-14-01 (17) | 91.7% | 91.2% — 17/17 |
  | 2026-06-15-02 (9)  | 88.1% | 85.0% — 9/9 |
  | 2026-05-21-01 (39) | 88.1% | 88.9% — 37/39 |
  | 2026-05-02 (49)    | 85.7% | 90.0% — 41/49 |

  Residual film tilt left in the Batch crop drops from a median of 0.15° to 0.01°, and
  frames rotated more than half a degree off their own film edge go from 14 to 3.
  
  ## What to try

  Run both Auto Crop and Batch Autocrop over a roll and look at the edges at full size.
  The interesting cases are camera scans that fill the frame, crooked or misaligned
  holders, sprocket rails, and thin (low-contrast) negatives. Flatbed scans with a wide
  rebate take the pre-existing path and should be unchanged — worth confirming.

  ## Known limits

  - A **wedge survives on sides the border crosses at an angle**. Clearing it means cutting
    the whole side back, which costs picture everywhere the border isn't, so that side
    deliberately abstains.
  - Batch **abstains on 8 of 49** frames on one roll, where the picture runs dark to the
    film edge and no peak stands out of the profile, and on 2 captures that each straddle
    two photographs. Those are left uncropped rather than cropped wrongly.
  - Two frames sit more than 2.5° from their roll's angle and are still straightened to the
    consensus, so they keep a wedge.

Fixes #828


seanharding:
@marcinz606 I'll let you use your judgement on how much testing you want on this before merging it in.

marcinz606:
I'll test a bit and merge as long as it's not worse :)

PR #829, opened 2026-08-14
Camera scanning: parse shutter labels that carry their unit (fixes #768)

### LINKED ISSUE #768: Nikon D600 Live Capture Preset Calibration Failure
Live capture on MacOS with a Nikon D600 and scanlight. Both the scanner and light are detected, but when I try amd calibrate a new film preset I get the following error. Camera is set to Manual.

“calibration failed: calibration: could not set shutter to '0.4': camera rejected it or it did not settle”

light-sntchr:
Fixes #768.

A Nikon D600 could not calibrate: `could not set shutter to '0.4': camera rejected it or it did not settle`. The camera captured and live-viewed fine, and its log showed the write was never refused, only ignored.

## Cause

The D600 publishes its shutter ladder as decimal seconds **with the unit attached**, straight from `gphoto2 --get-config shutterspeed`:

```
Readonly: 0
Choice: 24 0.0666s
Choice: 32 0.4000s
Choice: 36 1.0000s
```

`shutter_seconds` parsed neither form, so **every** label was dropped as junk, the body's ladder came out empty, and `normalize_start_point` fell back to the built-in one. That ladder is spelled the way a Sony spells it, so the calibration then wrote `"0.4"` onto a camera that has no such value. libgphoto2 accepts the string, the body ignores it, and the read-back never settles.

## Fix

The parser now strips a trailing unit (`0.4000s`, `2 sec`), leaving labels without one untouched and still dropping `Bulb`. With that, the D600's own 52-rung ladder survives: the third-stop spacing is read off it correctly and `true_seconds("0.3333s")` resolves to 2^(-5/3), so the physics is unchanged.

Two guards come with it, because the failure was silent for far too long:

- **A value the body never published is refused before the write**, and the log names what it does offer. Any future body with an unparseable ladder now fails visibly on the first attempt instead of producing "did not settle".
- **The shutter ladder respects the `writable` flag**, as the aperture already did, and calibration says so plainly rather than solving against speeds it cannot set. This one turned out not to be the D600's problem — it reports `Readonly: 0` — but the gap was real.

## Verification

New tests cover the unit-carrying labels, the real D600 ladder (spacing, `true_seconds`, and that the start point comes from the body's own vocabulary rather than ours), the refused-write path, and the read-only ladder. All of them fail without the change. `make all` is green with and without the optional camera group.

I do not have a D600, so confirmation from @sir-oscar-isaac-newton would be welcome.

### DIFF
--- a/negpy/desktop/view/sidebar/scanlight.py
+++ b/negpy/desktop/view/sidebar/scanlight.py
@@ -780,7 +780,13 @@ class ScanlightSidebar(QWidget):
         and meters noise instead of light."""
         data = self._settings_json()
         floor, ceiling = shutter_seconds(SHUTTER_CANDIDATES[0]), shutter_seconds(SHUTTER_CANDIDATES[-1])
-        labels = tuple(str(o.get("label", "")).strip() for o in (data.get("shutter") or {}).get("options", []))
+        info = data.get("shutter") or {}
+        if not info.get("writable", True):  # absent = unknown, so only an explicit read-only blocks
+            # A body that reports the shutter read-only will silently ignore every write, and the
+            # only symptom is a read-back that never settles (issue #768). Publishing no ladder is
+            # what makes the caller refuse instead of solving against speeds it cannot set.
+            return ()
+        labels = tuple(str(o.get("label", "")).strip() for o in info.get("options", []))
         # usable_ladder drops the unparseables (bulb-like "1/0", "Bulb", "") and returns ascending =
         # fastest-first; the range clamp on top is this UI's own policy, not the solver's.
         return tuple(label for label in usable_ladder(labels) if floor <= shutter_seconds(label) <= ceiling)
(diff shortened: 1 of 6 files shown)


MohammedAlkindi (commented):
Ran this on Windows 11 (Python 3.13, `uv`-synced env), branch `263c72a` against current `main` (`9c3f449`).

**Both touched test files, on the branch:**

```
tests/test_capture_calibration.py tests/test_capture_gphoto.py
88 passed in 4.92s
```

**Fail-before** — the PR's test files cherry-picked onto `main`'s source:

```
3 failed, 85 passed
  test_shutter_labels_may_carry_their_unit
  test_a_nikon_ladder_survives_and_keeps_its_own_labels
  test_a_value_the_body_never_published_is_refused_not_written
```

So the three new tests genuinely discriminate — they pin exactly the #768 mechanism (a Nikon's `0.4s`-style labels dying in a parser that only spoke bare numbers, and a fallback ladder being written to a body that never offered it) and the other 85 are unaffected either way.

Two things I like beyond the numbers: the `_choices` guard in `gphoto.py` turns the worst failure shape — a silently-ignored write that reads as "never settles" — into a refusal with the body's actual vocabulary in the log, which is the line a maintainer will actually get pasted in future issues; and the `writable` gate defaults open (`info.get("writable", True)`), so bodies that don't report writability keep today's behaviour.

Honest limits of this run: I have no gphoto2-compatible camera, so everything above is the mocked unit layer — the live Nikon path in #768 is untested by me; and the `scanlight.py` status-message branch (the "not reporting settable shutter speeds" UI path) isn't exercised by either test file, so I only read that part.

Issue #598, opened 2026-07-22
Colour profiles missing on exported tiff


andikulla:
Exporting tiff doesn't seem to include a colour profile while exporting png does.
My settings:
- Input files: 14 bit nef
- Export settings
  - Input ICC: none
  - Colour Space: P3 D65
  - Output ICC: none

The tiffs don't include neither a colour profile nor colour profile infos. Instead they include colour infos that match the sRGB ones according to `magick`:

```
Gamma: 0.454545
Chromaticity:
  red:   0.64 0.33
  green: 0.30 0.60
  blue:  0.15 0.06
```

The PNG exports do contain an icc profile `sP3`:
```
Profiles:
    Profile-icc: 480 bytes
```
and
```
icc:description: sP3
png:iCCP: chunk was found
```

I have to admit that I don't have the full picture about the colour space handling in negpy. It somehow makes sense to me that the input icc profile is `none` as I work with RAWs. But output is `none` as well with no chance to change it and I don't quite understand what's the difference between `Colour Space` and `Output ICC`. I would expect that `Colour Space` here means the actual internally used colour space while `Output ICC` is the colour space of the exported image and/or the included colour profile... Still, this seems to be an issue

andikulla:
Also, where's the 16 bit option? Or am I missing something?
EDIT: I guess removed and only available for flat exports intentionally. Probably because the "Print" export shouldn't need 16 bit as they are sort of done and shouldn't need 16 bit for minor edits... nonetheless I would sort of like to have an option for 16 bit print exports for extra editing headroom - but maybe that's just me 😌
EDIT 2: No, tiffs are 16 bit, png are 8 bit haha. So I can either get 16 bit tiffs without a colour profile but sRGB colour settings embedded or 8 bit png with colour profile (not tested any other format so far)
EDIT 3: JXL it is 😊 it's anyway the most modern format

thetalkingdrum:
I can confirm that there is a bug here. No matter which color space you choose, you get a TIFF without any color space tag. It also does not look like the preview in NegPy; it has different gamma and saturation. I tried several profiles in Photoshop, including my display profile, to see if I could get the TIFF to match the preview in NegPy, but I could not find a match. All the files look identical—changing the color space in negpy has no effect. Only when switching to Grayscale a color space tag written, and the image changes as well.

[...]

PR #833, opened 2026-08-15
fix(export): restore ICC color management on Flat/Print exports

thetalkingdrum:
## Summary

- Narrowband Scan's implicit ICC profile (and any custom Input/Output
  profile) was silently dropped on Flat exports whenever the source frame
  had an E-6-with-Normalize-off transparency transfer, because it was
  resolved using the pre-flatten (Print) `render_intent`. Fixed by
  resolving after `flat_master_config()`.
- Independently: on the officially released arm64 build, every export
  using Narrowband's implicit profile or a custom ICC profile silently
  no-opped color management — imagecodecs' lcms2 codec (`cms_transform`)
  failed to import. Root cause: PyInstaller bundles four separate copies of
  `liblcms2.2.dylib` (`cv2`, `PIL`, `rawpy`, `imagecodecs`) and collapses
  them into one canonical copy with no guarantee it's a compatible one —
  the arm64 build picked rawpy's, which is missing a symbol imagecodecs
  needs. Confirmed against the actual official 0.49.0 arm64 and x86_64
  DMGs: the x86_64 build happened to pick imagecodecs' own (complete) copy,
  which is why the original report saw this fail on M1 and work on Intel —
  not because Intel is safe from the collision, just a different pick.
  Fixed in `build.py` by repointing the canonical dylib at imagecodecs' own
  copy, with a permanent symbol-closure check (verified against all 7 real
  consumer files in the bundle) that fails the build if a future dependency
  bump breaks it again.
- Added a LUT-based fallback so a codec failure like this degrades
  gracefully instead of silently exporting unmanaged color, plus a
  diagnostic that surfaces the real underlying dlopen error in `negpy.log`
  if the codec ever fails to load again, instead of only a generic message.

Full investigation — including a debunked AMFI/notarization theory kept
intentionally as a record, the symbol-closure verification, and the DMG
comparison above — is in `LIBLCMS2_DYLIB_COLLISION.md`.

## Test plan

- [x] `make all` — clean except one pre-existing, unrelated,
      environment-sensitive test (`test_overflow_bar.py`)
- [x] Two new tests:
      `test_flat_export_keeps_narrowband_profile_despite_print_intent_transfer`,
      `test_cms_codec_unavailable_falls_back_to_lut_transform`
- [x] Verified end-to-end on real Apple Silicon (M1 Pro) hardware, across
      two separate rebuilds: `negpy.log` shows a clean ICC export with no
      fallback warning both times
- [x] Verified the dylib fix doesn't break the other consumers of the
      repointed library — symbol closure and runtime `dlopen` checked for
      all 7 real files across `PIL`, `rawpy`, `cv2`, `imagecodecs`
- [x] Verified against the actual official 0.49.0 release DMGs (not just
      wheels) which copy each architecture picked and why Intel wasn't
      affected


seanharding:
Heh, I was working on the liblcms2.2.dylib problem last night but didn't finish before I went to sleep.

marcinz606:
Interesting, pyinstaller picked different copy randomly on last build?

seanharding:
Was it just last build? I wasn't clear on when this problem appeared. I'm not 100% sure that the description is right. I went down this path because someone on Reddit claimed that the narrowband correction doesn't apply to TIFF exports. That didn't sound right to me, but I've mostly been exporting to JPEG lately, so I wasn't sure. In my testing, I found that JPEG exports seemed to work ok, but the color management wasn't working for TIFF exports (which did manifest as it looking like the narrowband correction wasn't working). I could clearly see that a JPEG looked different depending on whether I had narrowband on or off, but TIFFs looked identical regardless of that setting. I can also see that the JPEG does have the sRGB profile and the TIFF doesn't. The problem description here seems to imply that it applies to all exports, but that's not what I found. That said, it's the same root cause, so this should fix it.

thetalkingdrum:
Yeah, I was working on this during the past few days, but I didn't really have the time until today either :)

I think the bug might have been there for a while, but I didn't notice it until I got my M1 MBP back earlier this week. For some reason, it didn't manifest in the Intel Mac build, but that was just by "luck", so I started with the false assumption that it was working on the Intel build.

I think this might be the root of some of the hard-to-pin-down issues where a file was exported without an ICC profile on some systems but seemed to work on others, like in #598.

thetalkingdrum:
@marcinz606 I think the bug has been there at least since #598. I reproduced it then on my M1 MacBook Pro, but it was sent in for service a few days later. When I started working from my Intel Mac, the problem seemed to be gone, so I thought it might have been fixed somehow.

thetalkingdrum:
@seanharding You are probably right that the text is not true for all formats. I think I only tried tiff, ~~but PNG seems to also have been affected~~

marcinz606:
I never had that issue on any format on x86 Linux it must be arm64 mac specific

thetalkingdrum:
Claude seemed to imply that the problem was always there. But, at least on x86 macos, it would pick the right liblcms2.2.dylib by chance.

PR #868, opened 2026-08-16
fix(thumbnails): stop the segfault and the OOM when a folder's thumbnails generate

marcinz606:
## The crash

Loading a folder of scans sometimes kills the app:

```
Fatal Python error: Segmentation fault
Current thread ...
  File "PIL/Image.py", line 960 in frombytes
  File "PIL/Image.py", line 3458 in fromarray
  File "negpy/services/assets/thumbnails.py", line 151 in decode_source_image
  File "negpy/services/assets/thumbnails.py", line 229 in get_thumbnail_worker
```

## Root cause

Not a thread race — an out-of-bounds read.

`raw.extract_thumb()` on a Coolscan DNG returns `ThumbFormat.BITMAP` with an ndarray of shape `(166, 250, 3)` = 124,500 bytes. The file's embedded preview is **grayscale**: 166×250 = 41,500 bytes.

libraw hardcodes `colors = 3` for a bitmap thumbnail and reports the real payload length only in `data_size`. rawpy builds the ndarray from `(height, width, colors)` and exposes no `data_size`, so the array covers three times its allocation and PIL reads 83,000 bytes past the end.

Verified on `raw0001.dng`: the first 41,500 bytes equal tifffile's page-0 thumbnail exactly, and the rest is heap garbage (`... 113 171 2 0` — pointer bytes). It only faults when that tail meets an unmapped page, which is why it is intermittent and why four concurrent thumbnail workers make it likely.

The same unsafe read sat on the viewer's splash path (`preview_manager.py:103`), where `ensure_rgb` passed the `(h, w, 3)` array straight through.

## The fix

A BITMAP thumb's `data` is never read again. Both call sites go through one helper, `embedded_preview(raw, file_path)`:

- JPEG thumb → used as before (a real `bytes` copy, safe).
- BITMAP thumb → falls back to `_tiff_preview_page()`, which reads the file's own reduced-resolution page 0 with tifffile (guarded on SubIFDs present and uint8).
- no thumb, or any failure → `None`; the caller falls back to the half-size demosaic (filmstrip) or skips the splash (viewer).

Reading the preview page keeps the fast path for scanner DNGs rather than paying a demosaic: 2ms for a splash, against 168ms for `postprocess(half_size=True)` on that file.

## Tests

- New `tests/test_embedded_preview.py`. The BITMAP thumb's `.data` is a property that raises, so any future read fails the suite.
- `test_preview_manager_splash_grayscale.py` covered this case on a wrong premise — it asserted rawpy hands back `(H, W, 1)`, which rawpy never does. Rewritten to assert the buffer stays untouched and the splash comes from the preview page.
- `test_preview_manager_open_count.py`'s fake thumb switched to JPEG so it still exercises the splash.

## Verification

- 290 thumbnail generations over the 29 DNGs that crashed, 4-thread pool, 10 rounds: every call returned an image, no crash, 3.9s.
- Splash on the same file: `(166, 250, 3)` in 0.002s.
- `make all` green (4209 passed), `make format` clean.

Worth reporting upstream: rawpy shapes the thumb array from `colors` while libraw hardcodes it to 3.

---

# Second commit: the OOM on a 419-frame folder

Loading a 419-frame folder then died with exit 137. The kernel log names the OOM killer, at **13.9 GB RSS**:

```
Out of memory: Killed process 359607 (python3) total-vm:23544048kB, anon-rss:13890368kB
```

Two causes, both on the thumbnail worker.

**1. The preview page was rejected for being 16-bit.** These scanner DNGs lay out page 0 as `uint16` (256×172, minisblack) with the full-res data in a SubIFD. `_tiff_preview_page` accepted only `uint8`, so all 419 frames fell through to the half-size demosaic — where the first commit's BITMAP thumb had cost nothing. Both depths are now read, scaled to 8-bit. This one is a regression from the first commit; the second cause is not.

**2. `preview_positive` ran on the full-size decode.** Its only consumer is a 256px square, but it inverted a 2526×1698 image first: `uint8 -> float32`, the sRGB decode evaluating both `np.where` branches, log density, then three `np.percentile` passes. About a gigabyte of transients per worker, and the batch runs one worker per core (16 here). The image is now shrunk before the inversion. A/B on four frames from both folders: visually identical.

## Measurements

Peak RSS sampled every 20ms:

| | peak RSS | wall |
|---|---|---|
| before | OOM-killed at 13.9 GB | — |
| preview page fixed only | 3.80 GB (16 files, 4 workers) | 6.7s |
| both fixes | **1.41 GB** (419 files, 16 workers) | 65s |
| worst case, no embedded preview at all (forced) | 3.49 GB (64 files, 16 workers) | 8.3s |

The worst case is bounded rather than growing with the file count, so a library of raws with no usable preview no longer walks into the OOM killer either.

## Tests

- `test_16_bit_preview_page_is_scaled_down` — the uint16 page is read and scaled.
- `test_inversion_runs_on_the_shrunk_image` — spies on `preview_positive` and asserts it never sees more than `thumbnail_size` pixels.

`make all` green (4211 passed).

### DIFF
--- a/negpy/infrastructure/loaders/helpers.py
+++ b/negpy/infrastructure/loaders/helpers.py
@@ -89,20 +89,22 @@ def _tiff_preview_page(file_path: str) -> Optional[Image.Image]:
     """Reduced-resolution preview page of a TIFF-based raw, or None.
 
     Page 0 holds the preview only when the full-res data sits in SubIFDs, which is how
-    DNG writers lay it out.
+    DNG writers lay it out. Scanner DNGs write that page 16-bit, so both depths count.
     """
     try:
         import tifffile
 
         with tifffile.TiffFile(file_path) as tif:
             page = tif.pages[0]
-            if not page.pages or page.dtype != np.uint8:  # type: ignore[union-attr]
+            if not page.pages or page.dtype not in (np.uint8, np.uint16):  # type: ignore[union-attr]
                 return None
             arr = page.asarray()  # type: ignore[attr-defined]
     except Exception as e:
         logger.warning(f"TIFF preview page read failed for {file_path}: {e}")
         return None
 
+    if arr.dtype == np.uint16:
+        arr = (arr >> 8).astype(np.uint8)
     return Image.fromarray(ensure_rgb(arr))
 
 
--- a/negpy/infrastructure/loaders/helpers.py
+++ b/negpy/infrastructure/loaders/helpers.py
@@ -5,10 +5,11 @@ from typing import Any, Optional, Tuple
 
 import numpy as np
 import rawpy
-from PIL import ImageCms
+from PIL import Image, ImageCms
 
 from negpy.domain.models import ColorSpace
 from negpy.infrastructure.loaders.constants import SUPPORTED_RAW_EXTENSIONS
+from negpy.kernel.image.logic import ensure_rgb
 from negpy.kernel.system.logging import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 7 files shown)


marcinz606:
## Second commit: the OOM on a 419-frame folder

`make run` on a 419-frame folder was killed with exit 137. Kernel log confirms the OOM killer, at **13.9 GB RSS**:

```
Out of memory: Killed process 359607 (python3) total-vm:23544048kB, anon-rss:13890368kB
```

Two causes, both on the thumbnail worker.

**1. The preview page was rejected for being 16-bit.** These scanner DNGs lay out page 0 as `uint16` (256×172, minisblack) with the full-res data in a SubIFD. `_tiff_preview_page` accepted only `uint8`, so all 419 frames fell through to the half-size demosaic — where the first commit's BITMAP thumb had cost nothing. Both depths are now read, scaled to 8-bit. This regression came from the first commit; the second cause below did not.

**2. `preview_positive` ran on the full-size decode.** Its only consumer is a 256px square, but it inverted a 2526×1698 image first: `uint8 -> float32`, the sRGB decode evaluating both `np.where` branches, log density, then three `np.percentile` passes. About a gigabyte of transients per worker, and the batch runs one worker per core (16 here). The image is now shrunk before the inversion. A/B on four frames from both folders: visually identical.

## Measurements

Same 419 files, 16 workers, peak RSS sampled every 20ms:

| | peak RSS | wall |
|---|---|---|
| before | OOM-killed at 13.9 GB | — |
| preview page fixed only | 3.80 GB (16 files, 4 workers) | 6.7s |
| both fixes | **1.41 GB** (419 files, 16 workers) | 65s |
| worst case, no embedded preview at all (forced) | 3.49 GB (64 files, 16 workers) | 8.3s |

The worst case is bounded rather than growing with the file count, so a library of raws with no usable preview no longer walks into the OOM killer either.

## Tests

- `test_16_bit_preview_page_is_scaled_down` — the uint16 page is read and scaled.
- `test_inversion_runs_on_the_shrunk_image` — spies on `preview_positive` and asserts it never sees more than `thumbnail_size` pixels.

`make all` green (4211 passed).

PR #872, opened 2026-08-16
Add nix flake

activexray:
Adds a nix flake for a pure-deterministic builds for Linux and MacOS!

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,243 @@
+{
+  description = "NegPy — a tool for processing film negatives";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+
+    pyproject-nix = {
+      url = "github:pyproject-nix/pyproject.nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+
+    uv2nix = {
+      url = "github:pyproject-nix/uv2nix";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,243 @@
...
+      );
+
+    pythonSets = forAllSystems (system: mkPythonSet nixpkgs.legacyPackages.${system});
+
+    mkNegpy = pkgs: pythonSet: let
+      venv = pythonSet.mkVirtualEnv "negpy-env" workspace.deps.default;
+    in
+      pkgs.stdenv.mkDerivation {
+        pname = "negpy";
+        inherit (pythonSet.negpy) version;
```
Run1e: `pythonSet.negpy.version` is `0.0.0`, presumably because of how pyproject.toml fetches the version dynamically from the VERSION file? Not sure. Package output is `negpy-0.0.0`

Can be fixed with
```nix
          version = lib.trim (builtins.readFile ./VERSION);
```

Run1e (commented):
Only one small nitpick regarding the package name.

I don't have a lot of experience packaging python applications so I can't really comment that much on pyproject-nix/uv2nix, but LGTM

Hardware acceleration confirmed working with an AMD 9070XT

flake.nix
activexray: ah great point. fixed.

Run1e:
LGTM

PR #891, opened 2026-08-18
feat(macos): give NegPy a menu bar, with Window and Help

### LINKED ISSUE #890: macOS: add a menu bar
NegPy builds no menu bar on macOS, so the menus and key equivalents a Mac app is expected to have are missing.

What that costs today:

- `⌘M` does not minimize the window and `⌘W` does not close it. macOS takes both from the standard Window menu, so in an app without one they do nothing.
- Zoom has no equivalent at all.
- There is no window list, so with a live view or calibration window open there is no way to bring a particular window to the front from the menu bar.
- Help — the tour, the shortcut overlay, the update check — is reachable only from inside the app's own panels, not from where a Mac user looks for it.

Windows and Linux draw a menu bar inside the window, and NegPy has no use for a strip of chrome above the canvas there, so this is macOS only and should leave both other platforms unchanged.

seanharding:
AppKit takes Cmd+M and Cmd+W from the standard menu bar, so in an app that builds none they do nothing: NegPy could not be minimized or closed by keyboard, and Zoom was unreachable altogether. This is the smallest bar that fixes that, plus Help.

Window carries Minimize, Zoom, Close, Bring All to Front and the list of open windows — the main window plus any live view or calibration window — with a tick against the front one. Every command acts on the front window, whichever that is, not on the main window. Help gathers what was already scattered: the tour, the shortcut overlay and its editor, the Analysis guide and the update check, plus Report an Issue, which opens the tracker. That last one gets no shortcut registry entry: a binding would advertise it in the shortcut editor and the ? overlay on Windows and Linux, where nothing can run it.

There is no View menu and no Enter Full Screen item. AppKit adds one of its own to any menu titled View, so ours only ever appeared beneath it as a duplicate. Full screen stays the platform's: the green button, and the item macOS puts in the View menu NegPy will have later.

The bar has no parent, which is what makes it application-wide. A bar owned by the main window is only shown while that window is in front, so the menus blinked out whenever the live view or a floating panel took focus.

Items dispatch through ShortcutManager.action_for, so a menu item and its shortcut can never run different code, and a rebind reaches the menu. Only Cmd combinations become key equivalents: AppKit fires those before Qt's shortcut machinery sees the event, so a bare "?" for Keyboard Shortcuts would fire while the user types into the film-strip search box. The window keys stay out of the shortcut registry — they are platform commands, and a native key equivalent silently outranks a QShortcut, so a rebindable copy could only disagree with the menu. A test proves no registry default collides with one.

macOS only. install_mac_menus returns None elsewhere, so Windows and Linux keep their menu-bar-free layout with nothing visibly changed.

Fixes #890

### DIFF
--- a/docs/KEYBOARD.md
+++ b/docs/KEYBOARD.md
@@ -72,3 +72,14 @@ While a test strip or ring-around is up, `[` and `]` turn that proof's ladder in
 | `Mouse Wheel` | Zoom in / out (up to 400%); **Reverse Scroll Zoom** in the toolbar **⋯** menu flips the direction |
 | `Middle Click` + `Drag` | Pan zoomed image |
 | `Left Click` + `Drag` | Pan zoomed image (when no tool is active) |
+
+## Menu bar (macOS only)
+
+NegPy has a menu bar on macOS, with a **Window** and a **Help** menu. These keys come from it. They are platform window commands, not NegPy actions, so they are fixed and do not appear in the shortcut editor.
+
+| Key | Action |
+|-----|--------|
+| `⌘ + M` | Minimize the front window |
+| `⌘ + W` | Close the front window (closing the main window quits NegPy) |
+
+Every other menu item uses the binding listed above for its action, and shows it only when that binding uses `⌘`. A shortcut bound to a plain key still works from the keyboard; the menu just does not print it, because macOS gives a menu key priority over everything and a plain `?` would fire while you type into a text box.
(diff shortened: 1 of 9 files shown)


seanharding:
The menus, for reference (and for people who don't have macOS, so can't view them). The "Send python3 feedback" item is there because I'm running on a beta OS. It won't appear on production macOS builds.

Screenshot 2026-08-18 at 13 50 39

Screenshot 2026-08-18 at 13 52 47

PR #897, opened 2026-08-18
Round to nearest when quantising, instead of truncating

seanharding:
**Summary:** every 8- and 16-bit export was half a level too dark, because the float-to-int conversion truncated instead of rounding. Four lines of arithmetic; the two greyscale conversions in the same module had always rounded, so black-and-white exports were already correct while colour ones were not.

`_to_uint8_jit` and `_to_uint16_jit` cast with `np.uint8(v)`, which truncates toward zero. Every sample therefore lands on the level below whenever it falls between two: the quantisation error runs [-1, 0] instead of [-0.5, +0.5], which is a -0.5 level DC bias and twice the RMS error. Measured on random data: mean error -0.4999 levels, 50% of all pixels one level off, a systematic 0.196% darkening of full scale at 8 bits.

It lands on every 8-bit export -- JPEG, PNG, WebP, thumbnails -- and on `converters.py`, which is the canvas itself. At 16 bits the same half level is 0.0008% of scale and matters only in principle. Quantisation runs after `working_oetf_encode`, so the bias is in gamma-encoded space and is roughly uniform across the tone scale rather than concentrated anywhere.

That this was an oversight rather than a trade-off is settled by the same module: `_float_to_uint8_luma_jit` and its 16-bit twin have always done `v = lum * scale + 0.5`. Four near-identical loops, two rounding and two not, so a black-and-white export was correct while a colour one was half a level dark. Both now match `np.rint` exactly, verified across the range and at the endpoints, with NaN and out-of-range clamping unchanged.

`services/export/print.py` had its own bare cast and now calls the shared quantiser; a second implementation is how the two drifted apart in the first place, and a test now fails if one reappears there.

Three existing tests asserted the truncated values (127 for 0.5, 32767 for 0.5). They were encoding the bug; updated with the arithmetic spelled out.

Dithering is deliberately not included here. It addresses banding rather than bias, needs a decision about whether the canvas gets it too, and is worth landing separately.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+- Fix: **Exports are no longer half a level too dark.** Converting the finished float image to 8- or 16-bit truncated toward zero instead of rounding to nearest, so every sample landed on the level below whenever it fell between two — a systematic 0.196% darkening of full scale at 8-bit, with half of all pixels one level off, on every export and on the canvas itself. The two greyscale conversions in the same module had always rounded, so a black-and-white export was already correct while a colour one was not. Both now round, and the print-layout path uses the same shared conversion rather than a second copy of it. Re-exported files will differ from old ones by up to one level, which is the point.
+
 ## 0.52.0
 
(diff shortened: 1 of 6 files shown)


marcinz606:
nice catch!

PR #931, opened 2026-08-20
Camera scanning: give the calibration lock test a shutter ladder

light-sntchr:
`test_running_calibration_locks_the_window_inputs` has been failing on main since c4ca66a (#829).

That change made `_on_calibrate_new_preset` refuse to start when the body publishes no settable shutter speeds, so a Nikon D600 gets a clear message instead of a write it silently ignores. The bare sidebar fixture publishes no ladder, so the test returns at that gate and then asserts on a run that never started.

The test now publishes a ladder, which is what a real body does.

## Scope change

This PR originally also carried a metadata-free clipping detector for discussion #906. @thetalkingdrum showed that the version here, anchored on `values.max()`, misses the pile whenever noise and PRNU scatter photosites above it, including at the default centered calibration crosshair.

I checked it with my own sweep rather than take it on faith. Over 392 ROI positions on the reporter's green frame with real saturation above 5 percent, the version here misses 22 and the one in #932 misses 1. Every one of my misses sits at x = 0.5, which is where the light is brightest, so the most photosites are driven past saturation and the spread of their individual full wells is widest. It failed worst exactly where the clipping was deepest.

#932 carries the better detector, so the detection work belongs there and this PR is reduced to the test fix. The diagnosis that started it is in the discussion and in #932's description.

### DIFF
--- a/tests/test_scanlight_sidebar.py
+++ b/tests/test_scanlight_sidebar.py
@@ -231,7 +231,7 @@ def test_closing_running_calibration_waits_for_worker_terminal_signal():
     assert w._lv_target is w.lv_image
 
 
-def test_running_calibration_locks_the_window_inputs():
+def test_running_calibration_locks_the_window_inputs(monkeypatch):
     """A calibration meters a fixed base at a fixed ISO/aperture, so the film-stock name, the base
     ROI and the ISO/aperture must not change under it mid-run. They lock when it starts and unlock
     on a terminal outcome (here a cancel) so a failed run can be adjusted and retried."""
@@ -239,6 +239,8 @@ def test_running_calibration_locks_the_window_inputs():
     w._camera_verified = True
     w._light_verified = True
     w.calib_window.image._set_crosshair(0.5, 0.5)  # place the base patch so the run can start
+    # A run also needs the body's own shutter ladder; without one the start refuses (issue #768).
+    monkeypatch.setattr(w, "_settings_json", lambda: {"shutter": {"options": [{"label": "1/250"}, {"label": "1/60"}]}})
 
     w._on_calibrate_new_preset("Portra 400")
 


thetalkingdrum:
Nice diagnosis — the frozen-probe finding (32/34 runs pinned at 63114) is a good catch, and the shape-based idea is the right instinct: no white-level metadata should be required to see a photosite pileup for what it is.

I dug into `_plateau_clip_fraction` against the same D800 file (`260819_Frame023_G.nef`) and found the 100% green-channel number doesn't reproduce at every ROI position — specifically, it misses at the *default, centered* calibration crosshair (`_CROSSHAIR_FRAC = 0.012`, `_CROSSHAIR_ASPECT = 2.5`, ~89×221px / ~9,800 green photosites).

At that ROI: `values.max()` is a handful of noise-elevated photosites (shot noise, maybe a hot pixel) sitting ~50 raw counts above the real saturation pileup. The window in the fix is anchored to that max (`[top-8, top]` vs `[top-40, top-8)`), so it lands entirely in the sparse tail above the pile and misses the pile itself — `pile <= below`, returns `0.0`, on a patch that is in fact 100% saturated.

I swept ~625 plausible crosshair positions over this frame at the production ROI size: 410 have real saturation above 5%, and 17 of those — clustered right around the centered default — read `0.0` from `_plateau_clip_fraction`. The unit test doesn't catch this because the fake data pins clipped photosites to one exact value with nothing above it (`clip_at=15778`, zero noise), which is cleaner than real sensor behavior.

A fix that held up on that same 625-position sweep (0 misses, 0 false positives, including down to a 4-count noise sigma on a clean base): instead of anchoring on `values.max()`, scan the top ~256 counts for the densest bin (`np.bincount` over that window), and gate on that peak being meaningfully denser (~6×) than the run of bins immediately above it, rather than assuming the peak *is* the max.

One more small thing: the early `if white <= 0: return 0.0` gates the whole function, including your new metadata-free path — a body with no white level at all (exactly the case the shape-based detector is for) currently gets nothing from either check.

Happy to share the sweep script / test cases if useful, or point at a diff — up to you how you want to take it from here. We ended up folding a fix in this direction into our own working branch alongside the `camera_white_level_per_channel` ceiling fix, will likely open that as a separate PR referencing this one.

light-sntchr:
You are right, and it reproduces here.

At the default centered crosshair on Frame023_G, 100 percent of the green photosites sit at or above 15770 and 51.7 percent exactly on 15778, but my window is anchored on values.max() at 15825 and lands on the PRNU shoulder above the pile: 1339 photosites in the top band against 3413 in the band below, so it returns 0.0 on a fully saturated patch. My own check used a strip at the right edge where the pile happens to be single valued, which is what hid it.

I ran my own sweep rather than take it on faith. Over 392 positions with real saturation above 5 percent, mine misses 22 and yours misses 1. All of my misses sit at x = 0.5. That is where the light is brightest, so the most photosites are driven past saturation, the spread of their individual full wells is widest, and that shoulder is exactly what my window falls into. It fails worst where the clipping is deepest, which is the wrong way round.

Your density scan over the top counts is the correct generalisation, and you are right that the early white <= 0 return gates the metadata free path.

So #932 should own the detector. I have stripped this PR down to the failing test fix, which #932 does not carry.

PR #932, opened 2026-08-20
fix(capture): stop trusting a generic white level for ETTR clipping

thetalkingdrum:
## The bug

Discussion #906: narrowband camera-scan calibration was letting a genuinely clipped
channel pass as on-target, producing punchier/more-contrasty trichrome merges than a
white-light scan of the same film.

`raw_channel_clip_fraction` and `linear_demosaic` compared raw sensor values against
rawpy's generic `white_level` (a Nikon D800 publishes 16383), instead of the body's own
calibrated ceiling (`camera_white_level_per_channel` = 15311). Verified against the
reporter's own files: this let a fully saturated green channel read 0.000% clipped and
pass calibration's 0.2% clip-fraction budget.

## The fix

- `_linearity_limit()`: reads the camera's calibrated ceiling when available, falls back
  to the generic one, clamps a bogus calibrated value to the format's own ADC max.
- `_user_sat()`: the same limit, corrected for LibRaw's black-level subtraction, so
  `postprocess(user_sat=...)` scales correctly on any body (not just this D800, whose
  black level happens to be 0 and hid the issue during initial testing).
- `_plateau_clip_fraction()`: a second, independent, metadata-free check — finds
  saturation by density shape (the densest bin in a search window near the ROI's own
  ceiling) rather than assuming the ROI's maximum sample sits on the plateau, since real
  sensor noise routinely scatters a few photosites above it.

`raw_channel_clip_fraction` returns the two signals separately, budgeted differently
(update from review, see below) — the plateau detector, which finds true saturation from
the data's shape, keeps the strict `MAX_CLIP_FRACTION` (0.2%) budget; the linearity-limit
checks (raw and demosaiced) get a new, order-of-magnitude looser `MAX_LINEARITY_FRACTION`
(2%), since a calibrated linearity limit is often conservative and a clean channel's
ordinary tail past it isn't lost data the way a plateau is.

**Side effect worth calling out explicitly** (flagged by @light-sntchr below): the
black-level-corrected `user_sat` changes ETTR exposure targeting on *any* body that
publishes a calibrated ceiling below its ADC max, not only the D800 case this PR fixes.
On a Sony a7C II (white=16383, calibrated=15360, black=512) this scales every
measurement ~1.069x, landing the ETTR target lower in raw terms with more headroom — a
beneficial direction, but a distinct effect from the clipping-fraction fix, not merely a
side effect of it.

Related to #931, which independently diagnosed the same discussion and proposed a
metadata-free shape detector. We found its `values.max()`-anchored window misses the
real pileup at realistic calibration ROI sizes/positions (including the default centered
crosshair) when the ROI's noise ceiling sits above the true plateau — left a comment
there with details. This PR folds a corrected version of that idea in alongside the
ceiling fix.

## Verification

Against the reporter's real Nikon D800 files (`260819_Frame023_{R,G,B}.nef`), whole-frame
(not the production ROI, which reads more extreme). "Within budget" means the check would
let calibration proceed; "over budget" means it correctly aborts as clipped:

| channel | linearity fraction | plateau fraction |
|---|---|---|
| red | 0.550% — within budget (2%) | 0.145% — within budget (0.2%) |
| green | 33.4% — **over budget**, correctly rejected | 31.2% — **over budget**, correctly rejected |
| blue | 0.198% — within budget (2%) | 0.010% — within budget (0.2%) |

Blue's linearity reading (0.198%) used to sit right against the old shared 0.2% budget —
one noisy run from a false abort on a clean channel. It now passes comfortably. Green is
genuinely blown out and is correctly still rejected by both signals — this fix loosens
the budget only for clean channels near a conservative linearity limit, not for real
clipping, which is what the PR exists to catch in the first place.

`make all` clean. Full suite: 4562 passed, 2 pre-existing failures unrelated to this
change (missing optional `pyopticfilm` dependency for the Plustek flatbed backend,
present on `main` before this change too).

## Resolved from review

@light-sntchr's two points from the earlier review are both addressed: the `user_sat`
side effect is now called out explicitly above, and the linearity-limit/plateau
conflation is fixed by giving them separate budgets as described.

### DIFF
--- a/negpy/infrastructure/capture/raw_demosaic.py
+++ b/negpy/infrastructure/capture/raw_demosaic.py
@@ -14,7 +14,44 @@ its own verification); it is NOT covered here.
 
 from __future__ import annotations
 
+from typing import Optional, Sequence
+
 import numpy as np
+
+
+def _linearity_limit(raw, positions: Optional[Sequence[int]] = None) -> int:
+    """The camera's calibrated linear-range ceiling (raw ADC scale), not the format's generic max.
+
+    `white_level` is the format's theoretical max code value; `camera_white_level_per_channel` is
+    the manufacturer's own calibration of where the sensor stops responding linearly, often lower
+    (a Nikon D800: 15311 vs a generic 16383). Trusting the generic number lets already non-linear
+    or clipped photosites read as clean. Falls back to `white_level` where a body carries no
+    calibrated table. `positions` restricts the minimum to specific CFA channel indices; None
+    takes it across all of them.
+    """
+    generic = int(raw.white_level or 0)
+    per_channel = raw.camera_white_level_per_channel
+    if not per_channel:
+        return generic
+    values = [per_channel[j] for j in positions] if positions is not None else per_channel
+    calibrated = int(min(values))
+    # A calibrated limit can't legitimately exceed the format's own ADC ceiling; a body
+    # reporting one anyway is bad metadata, not a wider range.
+    return min(calibrated, generic) if generic > 0 else calibrated
+
+
+def _user_sat(raw) -> Optional[int]:
+    """`_linearity_limit`, converted to the post-black-subtraction scale `postprocess()` wants.
+
+    `user_sat` is compared after LibRaw subtracts the black level from every photosite, so the
+    raw-scale limit must be corrected by the same amount or every body with a non-zero black
+    level is silently under-scaled (invisible on a body whose black level happens to be 0).
+    """
+    limit = _linearity_limit(raw)
+    if limit <= 0:
+        return None
+    black = min(raw.black_level_per_channel or [0])
+    return max(1, limit - black)
 
 
 def linear_demosaic(path: str, half_size: bool = False) -> np.ndarray:
@@ -44,6 +81,11 @@ def linear_demosaic(path: str, half_size: bool = False) -> np.ndarray:
             # substitution and makes the demosaiced scale a fixed multiple of the raw counts, which is
             # what a meter measuring absolute light requires and what CLIP_CEILING assumes.
             adjust_maximum_thr=0.0,
+            # Pin the scale reference to the camera's calibrated linearity limit (see
+            # `_user_sat`), not LibRaw's generic ADC ceiling — otherwise a body whose real
+            # limit sits below the format max reports up to 65535 for photosites that are
+            # already non-linear or clipped, and CLIP_CEILING never sees it.
+            user_sat=_user_sat(raw),
             use_camera_wb=False,
             user_wb=[1, 1, 1, 1],
             output_bps=16,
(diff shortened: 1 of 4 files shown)


light-sntchr:
Tested this against my own rig, a Sony a7C II, which saturates at its published white level and so cannot reproduce the D800 bug at all. Your plateau detector and the calibrated limit both read at most 0.0033 percent there, and 0.00000 percent on a bare light frame with no film, which is the most uniform field a scanner can produce. No false positives from either path on that body.

Two things worth raising.

I could not reproduce your red 0.550 and blue 0.198 at first, then found where they come from. The D800 does not saturate at 15311. I took the densest bin in the top of each channel's histogram: red and blue pile up at 16383, green at 15778. So 15311 sits below the real saturation point of every channel, by 1072 counts on red and blue. The photosites between 15311 and the true pile are 0.4059 percent on red and 0.1876 percent on blue, which is essentially your two numbers. They are not clipped photosites, they are photosites in the upper part of a range the sensor is still using.

That matters because camera_white_level_per_channel is a linearity limit while MAX_CLIP_FRACTION is a clipping budget. Feeding one into the other puts a clean blue channel at 0.19 against a 0.2 threshold, so one noisy run aborts a calibration on a channel whose real saturation is 600 counts higher. Your own plateau detector finds the true pile from the data on all three channels and needs no such compromise, so it may be worth letting it decide clipping and giving the linearity limit its own budget if you want to keep it.

Second, user_sat. On my body (white 16383, calibrated 15360, black 512) it scales every measurement by 1.069. That is a behaviour change for every camera publishing a calibrated ceiling below the ADC max, not only for the broken case: the same scene now meters 7 percent higher, so ETTR lands at a different exposure. It works in a helpful direction, since the target then sits lower in raw terms and leaves more headroom, but it is a separate effect from the clipping fix and worth saying so out loud in the description.

The detector itself is clearly better than the one I had. Sweep numbers are in the comment on #931.

thetalkingdrum:
Both good catches, thanks — pushed a fix for both.

On the conflation: you're right, and your histogram numbers match what I'd independently found earlier (this D800's red pile at 16383 too, same generic ceiling). Split `raw_channel_clip_fraction` to return `(linearity_fraction, plateau_fraction)` instead of pre-combining them. The plateau detector keeps the strict 0.2% budget (`MAX_CLIP_FRACTION`); the linearity-limit checks — both the raw one and the demosaiced one, since its ceiling is scaled off the same calibrated limit via `user_sat` — get a new, 10x looser budget (`MAX_LINEARITY_FRACTION = 0.02`). On your blue reading specifically: 0.198% now passes comfortably against 2%, instead of sitting one noisy run away from the old 0.2% cliff. Green is still genuinely blown out and still correctly gets rejected by both signals — the loosened budget only helps clean channels sitting near a conservative linearity limit, it doesn't weaken real-clipping detection, which is the whole point of this PR.

On `user_sat`: added an explicit callout in the description — didn't want to bury a real behavior change as an implementation detail.

PR #935, opened 2026-08-20
perf: preview worst-case drags and export pipeline

marcinz606:
## Summary

Second perf pass (after #777). Normal drag/settle frames were already fast; this targets the pathological preview drags and the export path. All numbers from the 890M iGPU, 23.5MPx scan at 1600px preview unless noted.

| Scenario | Before | After |
|---|---|---|
| White/black-point, trim, hue-trim drag | 117 ms | 9 ms |
| Grade drag with a local mask | 48 ms | 24 ms |
| Clip / analysis-buffer drag (re-meters by design) | 133 ms | 54 ms |
| Export render + encode (tiled JPEG) | 1.25 + 0.47 s | 0.96 + 0.17 s |
| Export 57MPx | 2.45 + 1.19 s | 1.91 + 0.46 s |
| Export ARW RGB-triplet render | 5.65 s | 3.71 s |

## Preview

- Analysis cache key narrowed to the fields the meter reads; WP/BP offsets, per-channel trims and hue trim apply as uniforms and no longer bust it.
- Process/Sensor sliders send `readback_metrics=False` mid-drag (tone.py pattern) instead of paying settled-frame cost per tick.
- Local-mask raster cached grade-free; grade re-applies only the factor map at upload.
- Meter prefilter (replay + log10 + block-median) cached without the clip sliders; block median of 4 via strided min/max/sum.
- Export/metadata sidebar sync debounced; gear library and contact-sheet templates mtime-cached (was per-tick disk I/O on the UI thread).
- Duplicate histogram repaint per settled frame dropped.

## Export

- Exact matrix/TRC ICC kernel for named-space 8-bit exports, within 1 LSB of lcms on every bundled space (pinned by a test); lcms kept for custom/LUT profiles. JPEG bytes shift up to 1 LSB.
- TIFF/PNG metadata embedded at the first encode; the post-hoc rewrite decoded and re-compressed the full-res file.
- Batch export keeps the GPU texture pool; generation-based eviction bounds VRAM when dimensions change. Export passes source hashes, so multi-preset export of a non-tiled frame skips re-upload and re-metering.
- Triplet/HDR/stitch parts decode concurrently.
- Tiled loop overlaps each tile's readback wait with the next tile's dispatch.

## Verified

- `make all` green; parity suites (`test_gpu_tiled_parity`, `test_gpu_stage_skip`, `test_pipeline_parity`) pass.
- New tests: ICC kernel vs lcms delta, analysis-key invalidation matrix, first-encode metadata equivalence, batch cleanup contract.
- New CI metrics: `export.render_s`/`export.encode_s` plus worst-case drag metrics; render files added to the metrics workflow path filter.

## Rejected on the way

- Trilinear-LUT ICC transform: max delta 17/255 at gamut-clip creases.
- Demosaic-chooser "hotspot": profile artifact of `raw.raw_type` triggering libraw unpack, which postprocess needs anyway.

### DIFF
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -19,6 +19,8 @@ on:
       - "negpy/desktop/workers/render.py"
       - "negpy/desktop/prefetch_logic.py"
       - "negpy/kernel/caching/logic.py"
+      - "negpy/services/rendering/image_processor.py"
+      - "negpy/services/rendering/gpu_engine.py"
       - "tests/metrics/**"
   pull_request:
     branches: [main]
@@ -28,6 +30,8 @@ on:
       - "negpy/desktop/workers/render.py"
       - "negpy/desktop/prefetch_logic.py"
       - "negpy/kernel/caching/logic.py"
+      - "negpy/services/rendering/image_processor.py"
+      - "negpy/services/rendering/gpu_engine.py"
       - "tests/metrics/**"
 
 concurrency:
--- a/negpy/desktop/view/sidebar/process.py
+++ b/negpy/desktop/view/sidebar/process.py
@@ -279,10 +279,10 @@ class ProcessSidebar(BaseSidebar):
         self.sync_ui()
 
     def _on_white_point_changed(self, val: float, persist: bool = True) -> None:
-        self.update_config_section("process", persist=persist, **{self._wp_field(): val})
+        self.update_config_section("process", persist=persist, readback_metrics=persist, **{self._wp_field(): val})
 
     def _on_black_point_changed(self, val: float, persist: bool = True) -> None:
-        self.update_config_section("process", persist=persist, **{self._bp_field(): val})
+        self.update_config_section("process", persist=persist, readback_metrics=persist, **{self._bp_field(): val})
 
     def _on_lock_bounds_toggled(self, checked: bool) -> None:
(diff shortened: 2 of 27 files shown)


marcinz606:
Fixed a regression the dedup introduced: entering Peek Negative stopped switching the Analysis chart into the peek branch, because the peek paint emits image_updated only and the histogram refresh had moved to metrics_available. The refresh now rides image_updated, the one signal both paths emit (56930ea7).

Also added regression pins for the pass's contracts: drag readback flags, texture-pool eviction, gear/template mtime caches, block-median fast path (e667d51e).

marcinz606:
Valid catch: 6c9cd33a overwrote the pre-existing 370-line suite instead of appending. Restored verbatim from main in 3aeb9692 with the new TestMatrixTrcFastPath class appended; all 27 tests pass, including the imagecodecs cms ImportError fallback and the appearance-preservation check, both unchanged.

marcinz606:
dcaff374 fixes the "Dust Removal toggle does nothing until a manual heal" report. Not introduced by this branch — reproduced identically on main — but it rides the same mechanism this PR touches: the speck bake mutates the source without contributing to source_hash (hair_bake_token joins only when a hair is detected), so the engine's upload guard kept serving the stale texture. Latent since stage-skipping went live in #777. New luma_bake_token joins the hash in the preview, export and diptych constructions; pinned by a hash-wiring test.

PR #938, opened 2026-08-21
Improve handling of RGB triplets that don't sort in the correct order

### LINKED ISSUE #934: [Bug]: Loading a folder of trichromes leads to a mess of clashing photots
### NegPy version / build

51.1

### OS

MacOS 26.5

### Scanner

Digital camera (macro/copy setup)

### Scanner model / details

_No response_

### Light source

Scanlight v2

### Correction mode

On (narrowband trichrome correction enabled)

_No response_

_No response_

### What happened

It doesn't seem to matter the amount of photos in the folder but any more than converting one trichrome at a time leads to mismatches of photos and their colors

### Source file

n/a

### Screenshots

_No response_

seanharding:
The RGB triplet feature assumes that the files, when sorted by name, will be ordered such that captures from the same frame are next to each other. If they are not, the grouping fails.

For example, this works:

IMG_1234_B.NEF
IMG_1234_G.NEF
IMG_1234_R.NEF
IMG_1235_B.NEF
IMG_1235_G.NEF
IMG_1235_R.NEF

But simply renaming those exact same captures so they sort as follows will make it fail:

B_IMG_1234.NEF
B_IMG_1235.NEF
G_IMG_1234.NEF
G_IMG_1235.NEF
R_IMG_1234.NEF
R_IMG_1235.NEF

In this set of changes, we do two things to address this.

First, we sort by capture time before grouping, and only if that fails, fall back to sorting by filename. This too is imperfect — if the user captures in a weird order, or if the captures are too close together and the timestamp not granular enough (e.g. two captures on the same second), it can still fail. But it is much more robust and one of the two methods should work for most situations.

Second, we added a dialog box that pops up if RGB triplet mode is on and the files fail to merge correctly. This dialog attempts to give more concrete instructions on how the files need to be shot and organized for the feature to work. It gives a different message depending on whether the folder appears to be RGB triplet captures that just can't be combined for some reason, or if they don't look like RGB triplet captures at all.

I have another phase of this that groups based on file content, which works even if the filenames and timestamps are completely wrong, but that's a bit more error-prone and I think this should cover most cases, so I'm working on that separately.

Fixes #934


seanharding:
Oh, and since I didn't state it explicitly: this does correctly handle the example files from #934 — it would have prevented him from having to rename his files for them to merge.

PR #947, opened 2026-08-21
fix(preview): stop the layout pass upscaling the preview

seanharding:
1:1 zoomed closer than one scan pixel per device pixel whenever the decoded preview landed below preview_render_size. The layout pass takes render_size_ref as the paper long edge and resamples the content to fill it, so a 6016 px NEF, which decodes half-size to 3008, was upscaled to 4000 under a 4000 px preview. The canvas quotes zoom against render_long_edge, the buffer the pipeline was handed, so 1:1 read 4000/3008 = 133%. Toggling HQ then re-decoded at full resolution and 1:1 landed somewhere else.

The 1600 default hid it: PreviewManager always handed the engine exactly 1600 px. Cropped frames hit it at any preview size — a 4x crop was upscaled back to the long edge, and 1:1 there was really 400%.

Preview renders now re-derive the layout from the size the content can fill. The border keeps its share of the frame, since the virtual DPI scales with the same ref, and the display shader magnifies instead, bicubic when it does. Exports pass no size_ref and may still upscale to a print size.

Measured on a 6016 px NEF at preview_render_size 4000: zoom_level at 1:1 was 12.23 with HQ off and 9.20 with HQ on; both are 9.20 now.


seanharding:
Just realized I think something is wonky with this. Temporarily converting to draft until I nail it down.

seanharding:
False alarm. I was comparing between the release build and my local build but I hadn't correctly matched up all of the settings on the image I was looking at. This appears to work as I intended (tested with `preview_render_size = 4000`).

Issue #941, opened 2026-08-21
Metadata Export Presests


Icodextrin:
Combine existing gear presets into a metadata preset which contains all exportable metadata fields.

This will allow for easier selection of metadata on images such as scanner setup, developer, dev times, etc.

_Originally posted by @Icodextrin in https://github.com/marcinz606/NegPy/discussions/930_

Icodextrin:
closed with #950

PR #950, opened 2026-08-21
feat(metadata): metadata presets and a metadata library

Icodextrin:
Implements #941 

## What Changed

This pull request changes how gear presets work in NegPy. Previously the presets lived in their own library `~/NegPy/presets/*.json` which pointed at entries in `~/NegPy/gear/*.json`. It didn't cover any development or scanner info, and could not be saved via a preset.

Now presets are any set of metadata values, and the library added the two that were missing (Process and Scanning).
The flow looks like this now:

```
library record → MetadataConfig → payload → EXIF/XMP     (export)
                              ↘ facts_for → search index  (search)
```

I also added three new fields to the development tab: dilution, time, and temperature. Now you can search via these terms as well.

## UI Changes

Added new metadata preset tab:
image

New process tab:
image

New scanning tab:
image

New library preset:
image

When editing you may select what metadata gets included in the preset:
image


Icodextrin:
need to do some more testing / rewrite the claude generated pull request but this should be the bulk of the changes.
will update soon with screenshots of UI changes and a more clear explanation of how the new preset system works.

Icodextrin:
Fixed some lingering formatting issues, made some tweaks to the migration script that migrates the old presets to this new model, and tweaked a few of the default view options when creating a new preset.

marcinz606:
really nice!

PR #968, opened 2026-08-24
fix(export): stop Linear Output crashing on a camera DNG

Icodextrin:
## What

Exporting Linear Output from a camera DNG raises `ValueError: too many values to unpack (expected 3)`.

`_decode_linear`'s DNG branch unpacked `_decode_camera_raw`'s 4-tuple into 3 names:

```python
rgb, ir, wb = _decode_camera_raw(file_path, geometry)   # returns (f32, None, wb, meta)
```

**Who hits it:** anyone exporting Linear Output from an ordinary camera DNG (a DSLR-scanned negative, or Lightroom/DxO output). Not a *LinearRaw* DNG from VueScan/SilverFast — that takes the branch above, which works. `.dng` is in `SUPPORTED_RAW_EXTENSIONS`, so `_is_camera_raw()` matches it.

## Why the fix removes the branch instead of repairing the unpack

That early branch was a stripped-down duplicate of the general camera-RAW handler fifteen lines below it:

| | early DNG branch | general handler |
|---|---|---|
| RGB-triplet scans | ✗ | ✓ |
| Flat-field correction | ✗ | ✓ |
| Sensor matrix | ✗ | ✓ |
| White balance bake | ✗ | ✓ |
| TIFF + rawpy metadata merge | ✗ (TIFF only) | ✓ |

Removing the branch lets camera DNGs fall through to the handler that does all of it, which is what `docs/PIPELINE.md` already describes for camera RAW.

## Notes for the reviewer

- `ty` reports it as `invalid-assignment`, which the CI type-check step currently mutes via `--ignore`. Unmuting that list is separate work and deliberately not in this PR.

### DIFF
--- a/negpy/services/export/linear_output.py
+++ b/negpy/services/export/linear_output.py
@@ -366,14 +366,10 @@ def _decode_linear(
         rgb, ir = _decode_pakon(file_path, geometry, expansion=expansion)
         meta = _SourceMeta(make="Pakon", model=_pakon_spec_desc(file_path))
         return rgb, ir, None, meta
-    if _is_dng(file_path):
+    if _is_dng(file_path) and _is_linearraw_dng(file_path):
         meta = _read_source_meta_tiff(file_path)
-        if _is_linearraw_dng(file_path):
-            rgb, ir = _decode_dng(file_path, geometry, expansion=expansion)
-            return rgb, ir, None, meta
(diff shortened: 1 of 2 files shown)


Icodextrin:
should maybe think about unmuting the CI type-check. will probably save some headache in the future.

PR #970, opened 2026-08-24
Nikon Coolscan scanning through nkscan

marcinz606:
Adds a Nikon Coolscan backend built on the `nkscan` driver, next to the existing SANE one. It gets what coolscan3 cannot give us: a native IR pass, hardware eject, openICE dust removal, and frame rectangles measured off the film instead of guessed from a fixed pitch.

The scanner panel is now grouped (Film / Quality / Framing / Output) and the line above Scan says what pressing it will do: frame count, resolution, extra passes, estimated size. Leaving a preview dialog now stops the pass instead of leaving the unit busy until the next scan fails.

Tested on a Coolscan LS-50 with the SA-21 strip feeder: measured strips, IR, ICE (clean) full strip scans, eject


marcinz606:
needs more testing

PR #978, opened 2026-08-25
fix(plustek): restore Prescan crop x-mirroring for mirrored sensors

TobbyTravel:
## Summary

Fixes the Plustek **Prescan crop being horizontally displaced** ("correct crop size, offset one way") on mirror_x devices (OpticFilm 8200i SE / 8100 V2).

## Root cause — a regression from #958

- pyopticfilm flips the image left–right in `assemble()` (`mirror_x`) because the sensor reads right-to-left. The Prescan the user frames on is therefore the **mirror-corrected** image.
- But NegPy passed the crop `window` **unchanged** into TA/hardware space. The hardware crops by **sensor** x (pre-mirror) while the user drew by **display** x (post-mirror), so the returned ROI was reflected → right dimensions, displaced x.

The mirror-aware conversion **existed and was correct** since the Plustek backend shipped (#783) — `crop_to_scan_window(..., mirror_x=...)` with the x-reflection `(x1,y1,x2,y2) → (1-x2, y1, 1-x1, y2)`. **PR #958** ("feat(plustek): upgrade to pyopticfilm 1.1.2") silently replaced it with a plain `clamp_scan_area` at all three call sites and deleted the capability plumbing. This PR restores that behavior.

## Changes

- `negpy/infrastructure/scanners/params.py`: restore `crop_to_scan_window(crop, *, mirror_x)` — self-inverse x-reflection.
- `negpy/infrastructure/scanners/base.py` + `plustek_backend.py`: restore `prescan_mirror_x` capability, wired from `model.mirror_x`.
- `negpy/desktop/view/widgets/prescan_dialog.py`: apply the conversion at all three crop↔window points (default-crop display, window change, accept).
- Tests: x-reflection, passthrough, self-inverse; `prescan_mirror_x` surfaced on 8100 V2 caps.

## Validation

- Full suite: **4882 passed, 12 skipped**.
- **Hardware: verified on OpticFilm 8100 V2** — a Prescan crop now lands exactly where drawn.

## Scope

Plustek-backend only. `PrescanCropDialog` is gated on `capabilities.prescan`, which only Plustek sets; SANE/PieUSB use their own window handling and are unaffected. (Confirmed during investigation.)

### DIFF
--- a/negpy/desktop/view/widgets/prescan_dialog.py
+++ b/negpy/desktop/view/widgets/prescan_dialog.py
@@ -18,7 +18,7 @@ from negpy.desktop.converters import ImageConverter
 from negpy.desktop.view.widgets.scan_window_label import ScanWindowLabel
 from negpy.desktop.workers.scan_worker import PrescanRequest
 from negpy.infrastructure.scanners.base import ScannerDevice
-from negpy.infrastructure.scanners.params import clamp_scan_area
+from negpy.infrastructure.scanners.params import crop_to_scan_window
 from negpy.infrastructure.scanners.result import ScanResult
 
 
@@ -53,8 +53,10 @@ class PrescanCropDialog(QDialog):
         super().__init__(parent)
         self._controller = controller
         self._device = device
+        caps = device.capabilities
         # TA / backend space (what ScanParams.window stores).
         self._scan_window: tuple[float, float, float, float] | None = initial_window
+        self._prescan_mirror_x = bool(caps.prescan_mirror_x)
         self._busy = False
 
         self.setWindowTitle("Prescan — set crop")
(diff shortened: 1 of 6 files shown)


TobbyTravel:
Paired with the GL128 first-scan prime (jboneng/pyopticfilm#16) on the OpticFilm 8100 (V2): Prescan crops land exactly where drawn, stable across single-pass, Multi-Exposure and settings changes within one device session.

For review, the pairing has a versioning wrinkle: this PR is fully green against the current `pyopticfilm==1.2.0` release (lock) and mergeable as-is. But upstream pyopticfilm main has since renamed `bringup.is_opticfilm_8200i_se` → `is_gl128_opticfilm` (commit 92d513d, post-1.2.0, and widened to accept the 8100 V2 pid 0x1824 alongside the 8200i SE 0x1825). The next pyopticfilm release after the #16 merge will necessarily carry that rename, at which point the Plustek backend's imports break. Suggested follow-up (small, could also be folded in pre-merge): bump the `plustek` extra floor to `>=<next release>`, `uv lock`, and adopt the new symbol name — the rename is strictly more correct (it matches models by `asic == GL128`, with the SE pid kept for fallback), so no behaviour change for 8200i SE.

PR #989, opened 2026-08-28
Scan tab UI tweaks for scan tab when using pyOpticfilm backend

jboneng:
## Summary

Improves Scan tab feedback and layout for pyOpticfilm (and shared scan flows):

- **Prescan** on Plustek backends sits directly above **Scan** as a full-width control with crop icon; IR and Multi-exposure move to a spanning **Options** row when depth is hidden (8200i SE).
- **Progress bar lifecycle** is centralized: `set_scanning(False)` hides and resets the bar; indeterminate phases use the same pattern as other busy dialogs.
- **Post-scan busy states** cover the silent gaps after hardware finishes: **Merging exposures** (ME only), **Saving** (worker write), and **Importing scan…** (asset discovery until the frame is selected). Scan stays disabled until import completes.
- **`scan_import_finished`** signal wires discovery completion back to the sidebar so status shows the basename only after the frame lands.

## Test plan

- [x] OpticFilm 8200i SE: Prescan appears above Scan; full-window hint shows before crop is set
- [x] Single scan: progress moves through Scanning → Saving → Importing scan… → Scanned: `<basename>`
- [x] Multi-exposure scan: indeterminate **Merging exposures** appears after the last pass, before Saving
- [ ] Cancel / scan error clears progress bar and re-enables controls
- [x] `uv run pytest tests/test_scan_sidebar.py tests/scanners/test_plustek_backend.py tests/test_scan_worker.py -v`
- [x] `uv run ruff check`


jboneng:
Will mark as ready for review after #984 is merged into main

PR #1005, opened 2026-08-29
feat(export): one Export profile control, and an ICC import button

marcinz606:
Closes the confusion in #988.

`Color Space` and `Output ICC` both named the export destination, and the ICC silently won — at encode (`_encode_export`) and in the preview proof (`effective_output_icc`) — so the color-space combo stayed live-looking while doing nothing. That is why the thread went round in circles on "Soft Proof" / "Custom Export ICC" / "Device ICC": no pair of names describes two controls where one voids the other. This merges them.

### Export profile

One combo: the color spaces, then the imported ICC profiles.

```
COLOR MANAGEMENT                            [📁]
Processing is scene-linear (Adobe RGB primaries)
Input ICC        [ None                       ▾ ]
Export profile   [ Same as Source (Adobe RGB) ▾ ]
☑ Proof on screen (preview matches export)
```

`values()` still returns both stored keys, so `WorkspaceConfig`, `ExportPreset`, the encoder and the `{{ colorspace }}` filename variable are untouched. JXL greys custom profiles by the same rule as an untaggable space, which replaces the separate force-off-and-disable.

`Output ICC` is deliberately **not** renamed to "Soft Proof": it is converted to and embedded in the written file whatever the checkbox says, so that name would promise preview-only behaviour it does not have.

### The rest

- **`Same as Source` names the space it resolves to.** Showing nothing is what read as "NegPy ignored my profile" in the original report.
- **Import ICC** (the folder button on the section header) copies a `.icc`/`.icm` into `~/NegPy/icc/` and refreshes both lists with no restart. It never overwrites: different bytes under a taken name land beside it as `name (2).icc`, and a name matching a built-in space asks first, because `ColorSpaceRegistry` would then prefer it everywhere.
- **Soft proof and its warning move beside the profile they describe.** The warning sat inside the collapsed Preview section, so the one hint explaining a preview/export mismatch could not be seen while it fired. It now also fires for a custom ICC, not only a narrow space. `Display` stays in Preview — it is a per-machine setting, not part of the export recipe.
- **The HUD color pill is gone.** It read only the combo, never the ICC override, so it could name a space the file would not be in. The pill keeps the process mode.
- **A hint says what the pipeline actually works in.** Not "working space: Adobe RGB" — the pipeline is scene-linear with Adobe RGB primaries, and the Adobe TRC is applied only as the final engine step. `WORKING_COLOR_SPACE` names the boundary profile, not the internal state.

### Verification

`make all` green apart from the two pre-existing `pyopticfilm` failures on `main` (`tests/scanners/test_plustek_backend.py`, `tests/test_rgbscan.py`).

Nine new/rewritten tests: merged-combo round trip, stale-ICC fallback, the resolved `Same as Source` label, the JXL rules, and four for `import_icc_profile`.

Driven headlessly against the real `MainWindow` (16 checks): one destination control, the resolved label, the scene-linear hint, an import button matching the "Choose export folder" button pixel for pixel, the warning appearing and hiding with the toggle, an imported profile reaching both lists without a restart, and a HUD with no color pill.

### Open question

`Input ICC` keeps its name — nobody in #988 disputed that one. Say if it should change too.


thetalkingdrum:
Seems good to me.

Issue #738, opened 2026-08-02
NegPy crashes during export


P3RM-Mint:
Hello,
I’ve tried NegPy as an AppImage on LinuxMint (Linux Mint 22.3, kernel: Linux 6.8.0-136-generic) and am already very impressed with the automatic conversion when importing an image file, without the need for any further fine-tuning.
Unfortunately, the converted file cannot be exported.
After starting the export (Ctrl+E), a small export window opens; after a short while, the progress indicator freezes and, after some time, NegPy closes.
No exported image file is saved in the export folder (.../Documents/NegPy/...).
Furthermore, the available hard disk space decreases by approximately 600MB with every failed export attempt. I cannot tell which files are being stored where. Even if I delete the export path .../Documents/NegPy/..., nothing changes. Every new attempt to export uses up a further 600MB or so. After four attempts now, I’m missing about 2.4GB of storage space!

marcinz606:
can you launch appimage via terminal and say if there are any errors there while exporting?

P3RM-Mint:
Hello!

oliver@CF-20:~/Downloads/Software/NegPy$ ./NegPy-0.45.0-x86_64.AppImage
[2026-08-02 15:07:55] INFO     negpy.negpy.kernel.system.parallel: CPU parallel kernels: enabled (override=None, platform=linux)
Failed to create wl_display (No such file or directory)
Could not load the Qt platform plugin "wayland" in "" even though it was found.
[2026-08-02 15:07:56] INFO     negpy.negpy.infrastructure.storage.local_asset_store: LocalAssetStore initialized at /home/oliver/Dokumente/NegPy/cache
Forcing backend: Vulkan (6)
Unable to find extension: VK_EXT_physical_device_drm
[2026-08-02 15:07:57] INFO     negpy.negpy.infrastructure.gpu.device: WebGPU Backend: Intel(R) HD Graphics 615 (KBL GT2) (IntegratedGPU) via Vulkan
[2026-08-02 15:07:57] INFO     negpy.negpy.services.rendering.image_processor: ImageProcessor: Acceleration backend ready
[2026-08-02 15:07:57] INFO     negpy.negpy.services.rendering.image_processor: ImageProcessor: Acceleration backend ready
[2026-08-02 15:07:57] INFO     negpy.negpy.services.rendering.image_processor: ImageProcessor: Acceleration backend ready
[...]

PR #1016, opened 2026-08-30
fix: reduce tiled-export GPU memory pressure on integrated GPUs

TobbyTravel:
## Summary

Related to #738. #1006 fixed a wgpu-native panic-on-device-lost crash (an uncatchable abort across the FFI boundary) caused by handing an oversized buffer straight to the GPU on integrated GPUs -- but only for the HQ preview load path. #738's crash trace shows the same panic signature during export, through `_process_tiled`, on an Intel HD 615 iGPU.

`_process_tiled` already tiles large exports at a fixed 2048px, so there's no single oversized buffer the way #1006's preview crash had. What it does have: each tile's readback is deferred one tile behind its compute pass (an intentional throughput optimization), so two tiles' worth of intermediate textures/buffers are live on the GPU at once. On a device with a tight, unqueryable VRAM budget, that overlap is a plausible way to tip a submission over the edge.

This shrinks the tile to 1024px and drops the readback pipelining. **Originally this was gated on `GPUDevice.is_integrated`, but that lumps every integrated GPU together** -- a capable one (AMD's 890M, Apple Silicon) would take the smaller-tile/no-pipelining hit meant for older or memory-constrained hardware like the Intel HD 615 above, for no reason. Updated per review: it's now an explicit, **off-by-default** setting, `low_vram_export_tiling`, available in Preferences (Performance section, next to Multi-core CPU rendering) and as an `override.toml` key -- same override-wins-then-saved-setting-then-default resolution as `cpu_parallel`. The tiling/pipelining logic itself is unchanged, only what turns it on.

**This is untested on real hardware.** I have not reproduced #738's crash -- I don't have access to the reporter's setup, and the reasoning here is extended from #1006 (a confirmed fix, on confirmed hardware) rather than from a repro of this specific crash. Output is verified byte-for-byte identical to the default tiled path via new unit tests, so this should be a no-op for everyone until the setting is turned on. Since it's opt-in now, testing matters even more -- I'd appreciate it if @P3RM-Mint (or anyone else hitting export crashes on an integrated GPU) could flip the checkbox once this is available and report back before this is treated as resolving #738.

## Test plan
- [x] `make all` (format, lint, type check, full test suite) passes
- [x] New tests (`tests/test_gpu_tiled_low_vram.py`): `low_vram_export_tiling` uses smaller tiles; output parity with the default tiling (mean abs diff < 0.0005)
- [x] `override.toml` parsing/precedence and the Preferences checkbox covered in `tests/test_override_config.py` / `tests/test_preferences_dialog.py`
- [ ] **Not verified against #738** -- needs confirmation on integrated-GPU hardware, with the setting turned on, from the reporter or another affected user

### DIFF
--- a/negpy/desktop/main.py
+++ b/negpy/desktop/main.py
@@ -222,6 +222,12 @@ def main() -> None:
         # file's own apply() above, and still before QApplication reads any of them.
         apply_stored_override(override_cfg, APP_CONFIG, repo.get_global_setting)
 
+        # Opt-in smaller-tile/no-pipelining export path: override.toml wins if set, else the
+        # saved Preferences checkbox, else off. Not a STORED_PERF_KEYS entry since apply_stored
+        # skips bool values (see its docstring).
+        if override_cfg.low_vram_export_tiling is None:
+            APP_CONFIG.low_vram_export_tiling = bool(repo.get_global_setting("low_vram_export_tiling", False))
+
         # Multi-core Numba kernels: override.toml, then the saved setting, then the platform
         # default (off on macOS). Read before the flags below are overwritten.
         prev_clean_exit = bool(repo.get_global_setting("clean_shutdown", True))
--- a/negpy/desktop/view/widgets/preferences_dialog.py
+++ b/negpy/desktop/view/widgets/preferences_dialog.py
@@ -313,6 +313,27 @@ class PreferencesDialog(QDialog):
         grid.addWidget(hint_label(f"{os.cpu_count() or '?'} cores available. Applies at once, no restart."), row, 0, 1, 2)
         row += 1
 
+        self.low_vram_tiling_box = self._add_checkbox(
+            grid,
+            row,
+            "Reduce export memory use",
+            APP_CONFIG.low_vram_export_tiling,
+            "Use smaller tiles and less pipelining during export, at some cost to export speed. "
+            "Turn this on if exporting crashes on your GPU (typically an older or "
+            "memory-constrained integrated one).",
+        )
+        if "low_vram_export_tiling" in self._pinned:
+            self.low_vram_tiling_box.setEnabled(False)
+            self.low_vram_tiling_box.setToolTip("Set in override.toml, which wins over this dialog")
+        else:
+            self.low_vram_tiling_box.toggled.connect(self._on_low_vram_tiling_changed)
+        row += 1
+        note = "Applies to the next export started, no restart needed." + (
+            " Set in override.toml." if "low_vram_export_tiling" in self._pinned else ""
+        )
+        grid.addWidget(hint_label(note), row, 0, 1, 2)
+        row += 1
+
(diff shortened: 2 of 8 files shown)


marcinz606:
I think it's a bit excessive to do it on all `is_integrated` gpus. That way pretty decent ones like amd 890m i have in my thinkpad and also likely one in apple silicon would be gimped just to accomodate for some ancient intel iGPUs . Maybe instead this could be as a opt-in setting in `override.toml` and recently introduced preferences modal (default off)?

marcinz606:
20260830_09h42m17s_grim
here i mean

TobbyTravel:
Good catch — `is_integrated` isn't a reliable proxy for "actually memory-constrained." It would've cost export speed on your 890M or Apple Silicon for no reason, just because they're technically integrated.

Pushed a change: `is_integrated` gating is gone. The smaller-tile/no-pipelining behavior is now an explicit, **off-by-default** setting — `low_vram_export_tiling` — available in Preferences (Performance section, next to Multi-core CPU rendering) and as an `override.toml` key, following the same override-wins-then-saved-setting-then-default resolution as `cpu_parallel`. The tiling/pipelining logic itself hasn't changed, just what turns it on.

Since it's opt-in now, testing on real integrated-GPU hardware matters even more — @P3RM-Mint (or anyone else hitting export crashes on integrated graphics), if you get a chance to flip the checkbox and confirm it resolves the crash, that'd help before this is treated as fixing #738.

TobbyTravel:
Tried to reproduce #738 locally on my own integrated GPU (AMD Renoir/Lucienne, RADV/Vulkan — `is_integrated` reports `True`, 512MB dedicated VRAM carveout). Ran the app itself (clean launch, no crash) and also drove `_process_tiled` directly at a realistic ~63MP scan size with glow/halation maxed out to stress per-tile memory as hard as I could — both with the original full-pipelining path and with the new `low_vram_export_tiling` setting on. **Neither crashed.**

Watched `/sys/class/drm/*/device/mem_info_{vram,gtt}_used` during the run: VRAM stayed flat at baseline, but GTT (system RAM borrowed for graphics) spiked to ~5.7GB. This driver/GPU just overflows elastically into system RAM instead of hard-failing at the 512MB carveout — so it never hits the tight, unqueryable budget wall this fix targets.

So: I can't reproduce or rule out #738 from my end — my hardware is evidently a lot more forgiving than the reporter's Intel HD 615, which is itself supporting evidence that `is_integrated` alone was too blunt a signal (my machine would've been gated for nothing under the original approach).

I don't think this should be considered confirmed or ready to resolve #738 yet, even with the setting now opt-in — we still don't have a positive result showing the tiling/pipelining change actually prevents the crash on hardware that reproduces it. Still need @P3RM-Mint or another integrated-GPU user hitting the export crash to test with the setting enabled before treating this as a fix.

TobbyTravel:
Ready for Merge, as discussed in issue  #738

PR #1039, opened 2026-09-02
perf: host-side stalls around the GPU path (startup, navigation, export)

marcinz606:
## Summary

Profiled preview, export and loading with the GPU path as primary. The GPU frame was already 8–28 ms; the remaining latency was Python around it. Five wins plus one bug found on the way.

- **Startup**: pre-fill wgpu-py's cffi type cache from the parsed webgpu.h models, so window construction and the first render stop re-parsing ~60 type strings through pycparser.
- **Navigation**: small numba calls (chart, wedge) take the serial variant below `SERIAL_MAX_ELEMENTS` instead of waiting on the parallel gate behind the render thread; gear combos reload only when the library or selection changed; EXIF and profile TOML reads cached per (path, mtime); no `gc.collect` on file switch.
- **Export**: tile RGB↔RGBA copies in numba kernels; lcms 16-bit transform in row strips on a thread pool; tiled path honours `readback_metrics`. Output bytes identical.
- **Fix**: `gamut_lut` round-tripped into the input profile; with Narrowband Scan (`RGBScan.icc`, no B2A) every settle logged "cannot build transform" and the printability read-out stayed off.

## Measurements (same-condition A/B, 890M iGPU)

| Path | Before | After |
|---|---|---|
| MainWindow construction | 836 ms | 631 ms |
| First engine render | 0.8 s | 0.4 s |
| Open first RAW → painted | 1576 ms | 1250 ms |
| Navigation, prefetched buffer | 170–188 ms | 45–115 ms |
| GUI thread `select_file` per nav | 52 ms | 10 ms |
| Analysis panel per settle | 26 ms | 2 ms |
| 57 MP TIFF export render (warm) | 1160 ms | 946 ms |
| 57 MP TIFF16 encode | 1825 ms | 755 ms |

## Test plan

- `make all` green (5391 passed).
- New tests: cffi warm identity + parse count, RGBA kernels vs numpy, lcms strips vs single call, dispatcher size routing, TOML/EXIF cache invalidation, gamut LUT with an input-class source profile.
- Export bytes cmp'd before/after on a 57 MP TIFF (JPEG and TIFF16): identical SHA.

### DIFF
--- a/negpy/desktop/view/sidebar/metadata.py
+++ b/negpy/desktop/view/sidebar/metadata.py
@@ -477,8 +477,18 @@ class MetadataSidebar(BaseSidebar):
         self._gear_library = GearProfiles.load_library()
         library = self._gear_library
 
+        # Rebuilding five combo models per file switch is the panel's main sync cost, so a combo
+        # is only reloaded when the library or its selection changed.
+        seen = self.__dict__.setdefault("_gear_combo_keys", {})
+
         def should_refresh(combo: SearchableGearCombo) -> bool:
-            return force or not combo.is_editing()
+            if not force and combo.is_editing():
+                return False
+            key = (id(library), self._gear_selected_id(combo, conf))
+            if not force and seen.get(id(combo)) == key:
+                return False
+            seen[id(combo)] = key
+            return True
 
         if should_refresh(self.camera_combo):
             self.camera_combo.set_gear_items(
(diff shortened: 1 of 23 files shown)


marcinz606:
Added `5f36ca76` — dust removal at full resolution (HQ preview settle and export), measured on a 23.5 MP scan with an IR plane:

| Path | Before | After |
|---|---|---|
| IR OpenICE, first HQ render | 6.2 s | 1.4 s |
| IR NegPy method, first HQ render | 5.8 s | 3.4 s |
| IR threshold change at HQ | 5.7 s | 1.5 s (OpenICE) / 2.3 s (NegPy) |
| Export render with IR | 6.2 s | 1.6 s / 3.3 s |
| Optical dust at 57 MP export | +0.5 s | +0.2 s |

OpenICE: band math fused into one numba pass (output within float rounding, routing identical). NegPy method: fused fill/blend/floor passes (bit-identical), sparse grain donor lookup, hair inpaint no longer treats a frame-spanning component as a 22 MP crop, ratio/gain cache survives threshold drags.

marcinz606:
Added 07308368: export now reuses the HQ preview's dust bakes (export render after an HQ preview 3.8 s → 1.4 s at 23.5 MP with IR + optical dust), scratch tracing/scoring computes the ridge on a row band (hover 0.4 s → 0.09 s on a full-res buffer, bit-identical), brush strokes take density per crop.

Issue #65, opened 2026-01-28
Iconsistency when batch exporting


maltklaus:
Another thing I immediately ran into on the latest version:

I edited an entire roll and set the export settings to 40cm and 300 dpi. However, when selecting to export all loaded pictures, it didn't apply the same setting to all the pictures, some were much smaller.
When exporing the images individually, it worked as intended.

marcinz606:
per-file settings are still applied when running batch, i will add checkbox to apply same export settings to all when using batch in new version

PR #1041, opened 2026-09-02
feat(plustek): add Multi-Pass ME scanning

TobbyTravel:
## Highlights

- **New: Multi-Pass mode.** Merge 2-9 exposures per frame for lower shadow
  noise, at a roughly proportional cost in scan time. pyopticfilm's own
  benchmark on an 8100 V2 (5 exposures vs. today's 2, same frame): **-26%
  relative chroma noise, -28% relative luma noise**, visibly smoother in flat
  regions.
- **Clearer control.** The old "Multi-exposure" checkbox — which secretly
  only ever meant one specific behavior — is now a single dropdown: **Off /
  Dynamic (2 exposures) / Fixed (2 exposures) / Multi-Pass (2-9)**. Choosing
  a mode is now an explicit decision instead of a hidden default.
- **Tidier scan setup.** Frame selection, scan-window/Prescan, and Clear now
  live together as one full-width block right above the Scan button — the
  last thing you touch before scanning, grouped where you touch it. Crop
  info, when a crop is set, shows next to Frame info instead of its own row.
- **No new risk surface.** Every mode is a name for a fully validated
  pipeline path — there's no raw exposure field, no way to type in an unsafe
  value. The debug-only raw exposure overrides pyopticfilm keeps for its own
  ScanLab tooling are never called from NegPy.
- **Fully backward compatible.** Off stays the default; Dynamic at 2
  exposures is byte-identical to today's only multi-exposure behavior. No
  existing scan or saved setting changes.

## Technical changes in NegPy

- `ScanParams`/`ScannerSettings` gain a `MultiExposureMode` enum replacing
  `multi_exposure: bool`, plus a bracket count meaningful only in Multi-Pass.
  Old settings blobs migrate automatically (`multi_exposure: true` →
  `dynamic`, `false` → `off`).
- `plustek_backend.py` translates the mode into pyopticfilm's
  `scan(multi_exposure=, me_exposure_mode=, n_brackets=)`, and generalizes
  the previously 2-pass-only scan progress model to N passes.
- `ScanSidebar`: mode combo + bracket slider replace the checkbox; the
  Framing section (frame spec, scan window, Prescan) moves to a full-width
  block above the Scan button; crop info relocates next to Frame info.

## Why this is a PR draft

This depends on `pyopticfilm` API surface (`n_brackets`, `me_exposure_mode`
widened to `str | None`, `Model.me_default_exposure_mode`) that only exists
on pyopticfilm's `feat/me-n-brackets` branch (#47), not in any released
version yet. **This will stay a draft until that pyopticfilm work is merged
and released**, since it can't be installed as a normal dependency or fully
verified against a real release until then.

### DIFF
--- a/negpy/desktop/view/sidebar/scan.py
+++ b/negpy/desktop/view/sidebar/scan.py
@@ -21,9 +21,38 @@ from negpy.desktop.view.sidebar.base import install_wheel_guards
 from negpy.desktop.view.styles.templates import StatusStrip, hint_label, icon_button as _icon_button, section_subheader
 from negpy.desktop.view.styles.theme import THEME
 from negpy.infrastructure.scanners.base import ScannerCapabilities, ScannerDevice
-from negpy.infrastructure.scanners.params import FILM_TYPES, FilmType, film_passes_infrared
+from negpy.infrastructure.scanners.params import (
+    FILM_TYPES,
+    MAX_ME_BRACKETS,
+    MIN_ME_BRACKETS,
+    FilmType,
+    MultiExposureMode,
+    film_passes_infrared,
+)
 from negpy.infrastructure.scanners.registry import DEFAULT_BACKEND_ID, backend_choices
 from negpy.infrastructure.scanners.settings import ScannerSettings
+
+#: Label + explanatory tooltip for each multi-exposure mode, in display order. The exposure
+#: count is spelled out in the label itself (not just the tooltip) so Dynamic/Fixed's shared
+#: 2-exposure behavior — as opposed to More exposures' 2-9 — is visible without hovering.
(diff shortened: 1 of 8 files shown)


TobbyTravel:
Closing in favor of #1064.

This branch was built against an earlier pyopticfilm design (N different exposures merged, like HDR bracketing) that never shipped. pyopticfilm instead shipped [PR #65](https://github.com/jboneng/pyopticfilm/pull/65) with a different mechanism — repeating a single exposure `n_passes` times and stacking the aligned repeats for an SNR gain, orthogonal to the existing ME short+long fusion rather than an alternative to it (`n_brackets` isn't part of that API). #1064 reworks the UI/backend wiring for that actual API.

PR #1042, opened 2026-09-02
Per-frame registration for the strip preview, mono TIFF, and no more DNG output

marcinz606:
Three rounds of work on the strip preview and the scan writer, on top of the nkscan backend.

## Per-frame offset

Offset and Drift correct a strip whose gaps creep uniformly. Neither reaches a single frame that sits off on its own — a splice, a mis-fired advance, a boundary the detector placed a hair late. Each tile now carries its own slider, adding a correction on top of that ramp:

```
offset(N) = frame_offset_mm + (N-1) * drift + frame_offsets[N]
```

An absent key is 0.0, so nothing existing changes behaviour. No backend work was needed: the correction rides `frame_offset_mm`, which already slides the nkscan rect and writes the coolscan3 `subframe` option. A measured strip re-addresses an absolute rect and may go either way; a feeder clamps the total as before, because it cannot back up.

**The bug this PR also fixes.** The correction reached the scan only while its tile existed. `frame_offsets()` read the sliders and nothing else, and a measured strip opens with no tiles — they appear when the strip is detected. Reopening the preview and accepting without detecting answered with an empty map and wiped every saved correction, so the batch scanned unshifted. A frame with no tile now keeps what was saved.

The arithmetic was correct the whole way down, and the tests covering it all passed. They shared one blind spot: every dialog test built a device with an adapter capacity — a feeder, whose tiles are built eagerly — so nothing ever constructed a transport that measures the strip. There is a `_discovery_device()` helper now, and the end-to-end test drives real sidebar values instead of a hand-built request.

## Tile size, and something to judge the offset against

Tiles get a size slider (90–340 px, remembered). The label letterboxes its kept pixmap at paint time, so resizing costs no scanning. The grid no longer wraps at a fixed six — columns follow the available width, which is what keeps a large tile on screen — and it reflows when the dialog is resized.

Every previewed frame gets a thin outline. That box is the boundary the offset is measured from, and without it there was nothing to judge against on a dark scan. Drawn last, so neither the shaded band nor a crop taken to the edge dims it. It lives in the shared label, so the prescan and quick-scan previews get it too.

## Scan output: TIFF only, colour or mono

Film carrying a single record reads the same off one plane as off three, so Format gains **TIFF (mono)**: one 16-bit grey plane at a third of the size. The plane is the mean of the three, summed in uint32 so nothing overflows and rounded rather than floored — mono film meters with its channels locked, so the three are one density read three times. Derived host-side rather than asked of the transport, because nkscan offers no grey mode; the loader restacks a 2-plane TIFF, so NegPy reads its own mono scans back unchanged.

**DNG output is retired.** `write_dng_linear` and its encoder are gone, and a saved `"DNG"` preference lands on TIFF — the same mapping `RETIRED_EXPORT_FORMATS` already applies on the export side. Reading LinearRaw DNGs from SilverFast and VueScan is untouched; only writing them is dropped.

## Verification

`make all` green: format, lint, ty, and **5417 passed, 13 skipped**.

The per-frame chain was checked numerically rather than by reading, at every hop. With base 1.0 mm, drift +0.2 mm/frame and −0.7 mm on frame 3 alone, the dialog, the preview request and the batch all produce `1.0, 1.2, 0.7, 1.6, 1.8` mm, and the rect handed to nkscan shifts by exactly 110 stage units (0.7 mm at 4000 dpi) on the feed axis only, with the frame's length preserved and its neighbours untouched.

## Known limitation

The per-tile slider spans ±2.5 mm on every transport, but a feeder clamps any negative total to zero — it cannot back up. Consistent with how Drift already behaves, and the clamp notice says so, but a negative correction is only genuinely useful on a measured strip.


activexray:
nkscan does have a greyscale mode, but only through the ICC output path as nikon scan had some profile for this already that did RGB->Greyscale.

marcinz606:
> nkscan does have a greyscale mode, but only through the ICC output path as nikon scan had some profile for this already that did RGB->Greyscale.

i want something generic for all scanner backends, i think just averaging rgb reads that claude suggested makes a lot of sense

activexray:
I fixed a zillion little bugs with the USB transport so hopefully this is more stable. Next release i'll once again rework the frame detection

thetalkingdrum:
I tried this + nkscan 0.10 earlier today on an LS-50, and it would let me set the individual offsets. But when it actually did the scan, it seemed to just ignore them.

I didn’t have much time to look into it, but hopefully I’ll have some time to do a few more scans soon.

marcinz606:
Yeah, something is a bit off, not yet sure what. Offset is applied but it seems less than requested (or reported detected frame has some mismatch vs what I draw in strip preview). It needs some tweaking. When extreme per-frame offset is applied it is certainly there 

(this is with frame offset dragged all the way)

marcinz606:
I think i found the issue

thetalkingdrum:
The image link is broken, so I can’t see it. I also made the frame offset deliberately large just to see if it was working, but when I scanned it, it wasn’t.

marcinz606:
20260905_19h51m15s_grim

marcinz606:
this is max per frame offset, i think I've made it slightly better but still something is a bit off :|

thetalkingdrum:
ah yes, I also got one strange frame like that. But without setting to max as I remember it.. But on the other frames it was like nothing happened. I might have time to do some scans tomorrow.

thetalkingdrum:
Ok, did some new scans with the latest commit from 15 min ago. I’m still getting weird results. It’s like it’s not physically repositioning the film, but only telling the scanner to read further back in the mask or something like that.

thetalkingdrum:
Screenshot 2026-09-06 at 13 10 21
Screenshot 2026-09-06 at 13 14 02

thetalkingdrum:
Screenshot 2026-09-06 at 13 21 49
Screenshot 2026-09-06 at 13 24 40

thetalkingdrum:
Like in the last example, we should see parts of frame 4 entering in the scan of frame 3, as shown in the preview. Instead, it seems like the film isn’t being repositioned, and the stepping motor is simply being told to read further out. We see the SA-21 mask (where the black narrows in on the right side) and eventually the motor reaches its endpoint and then just scans that same line over and over.

That is my quick take.

marcinz606:
I think i narrowed it down to nkscan behaviour and fix belongs there

marcinz606:
https://github.com/activexray/nkscan/pull/53

marcinz606:
@thetalkingdrum can you verify with nkscan python wheels built from https://github.com/marcinz606/nkscan/tree/fix/perforation-lead ?

thetalkingdrum:
I’ll try later tonight

thetalkingdrum:
Screenshot 2026-09-06 at 21 57 47

thetalkingdrum:
Success!

activexray:
Can y'all try with the wheel from the tip of main? Trying to clean things up a bit.

thetalkingdrum:
@activexray Works:)  

just needed dropping the frames= arg locally to match the new scan_frame() signature

activexray:
Gonna make a few more tweaks to the python API btw

thetalkingdrum:
Correction. Turns out no mod was needed.

 NegPy's scan_frame() wrapper already checks inspect.signature(session.scan_frame) and only forwards frames= if the installed build actually accepts it, so it silently no-ops against your new API. Ran it unmodified just now, works fine. 

But found another bug or regression. Not sure what happened here but seems like it added the offset in the wrong direction. This is with the current main on nkscan:
Screenshot 2026-09-07 at 17 12 39

thetalkingdrum:
OK, something even weirder happened. I scanned the whole strip, but instead it scanned the fist frame six times: 

Screenshot 2026-09-07 at 18 05 43

marcinz606:
this is against current main nkscan, not my fork branch?

thetalkingdrum:
yes current main nkscan

thetalkingdrum:
Since @activexray asked us to try it out. I scanned one frame last night and that was ok. There were some commits after that and today it has been acting strange.

activexray:
I'll take a look yeah I'm still messin with things a bit. I'm trying to detangle frame stuff to match better what nikon scan does.

thetalkingdrum:
Ok, yeah I'm getting unpredictable results now, so that makes sense. Tried to reproduce the "same frame six times" bug, but this time it worked correctly, but the framing was off.

activexray:
unpredictable results on the same commit? that's bizzare, i don't think anything is non-deterministic

thetalkingdrum:
Idk. Maybe it’s the exact order of doing something. Or maybe because I’m
running into several bugs. But I couldn’t reproduce the bug where it
scanned the first frame six time instead of scanning the strip. And I
didn’t have time to look into the one where it would set the offset in the
wrong direction.

man. 7. sep. 2026 kl. 20:14 skrev Kira ***@***.***>:

activexray:
alright, please give a try now. I've reworked frame detection (again) and I think I haven't mucked up the previous fix. There are a few more tweaks to the python api.

marcinz606:
I'll test it out tomorrow and if it's good then hopefully finalize this PR

thetalkingdrum:
Did some quick tests before heading to work. I had claude patch negpy to accommodate the latest changes. But I was now back to the original problem where an offset would not be honored. 

Also there seemed to be a mismatch between the preview and the scan. The preview did have some badly framed frames, but the scan came back correctly framed without touching the offset. 

But I really didn’t have enough time to make any conclusions. I’ll let Marcin look into it. Maybe I’m doing something wrong.

thetalkingdrum:
Btw @activexray would it be possible to add nkscan's logging to the wheel? Seems like tracing just isn't hooked up for the python feature, only the CLI. Feels like it should be easier to debug if we have both logs.
The issues here would probably be easier to track down with it. Also flagging this one, if you get a chance to take a look.  I have a hunch nkscan logs could tell us more there too: https://github.com/marcinz606/NegPy/issues/1066

thetalkingdrum:
Apparently I was only on 6094418 when testing this morning. Rebuilt with 0.11 now but still same result: tried stetting a large offset, but the frame comes back with no offset.

activexray:
Alright could you try again with main, I added an `init_logging` to python where you can run `init_logging('trace')`

activexray:
> Also there seemed to be a mismatch between the preview and the scan. The preview did have some badly framed frames, but the scan came back correctly framed without touching the offset.

I might need to dig into what negpy means by preview. I don't have a separate "preview" pass from "scan" or "thumbnail". There is one in Nikon Scan, I just haven't wired it up yet. afaik it's just a lower DPI pass.

thetalkingdrum:
what i asked for: Screenshot 2026-09-09 at 17 56 15
what i got:
Screenshot 2026-09-09 at 17 59 59

[negpy.log](https://github.com/user-attachments/files/32016599/negpy.log)
[nkscan_trace.log](https://github.com/user-attachments/files/32016600/nkscan_trace.log)

thetalkingdrum:
claude's diagnosis: A per-frame offset is silently ignored on main (0.11.0 and the 9f855bc build with init_logging). Tested through both raw nkscan.Session calls and through NegPy's own UI; same result either way.

Repro: discover a strip, then request a whole frame shifted along the feed axis (e.g. frame 4's detected rect, top moved by a few mm). The scan comes back — correctly cropped, frame_columns well-formed — but it's the original frame's content, not the shifted one. Confirmed by pixel correlation against a direct unshifted scan of the same frame (r≈0.97-0.99) across several repeats and offset magnitudes.

With init_logging("trace") wired into NegPy, one scan (frame 4, -8.30mm, requested top 16901 vs detected 18208) shows the whole story in three lines:

DEBUG registered a frame off the table top=16901 line=404 perf=PerforationInformation { perf_number: 23, ... }
DEBUG registered a frame off the table top=18128 line=433 perf=PerforationInformation { perf_number: 24, ... }
DEBUG registered the frame where the metering pass measured it from=16901 to=18128

So: framing::register does write a genuine new table entry for the shifted top (16901) — that part works. But nothing between that write and the metering pass (focus_frame, then autoexpose_frame_with in scan_frame_with, src/scan/frame.rs) makes the unit actually seek to the newly-registered address first. The metering pass measures the picture still starting near the frame's original position (18128, close to the detected 18208), and recentered() reads that as "the frame needs correcting" and calls register() again — this time at 18128 — which overwrites the table entry we just wrote for 16901. End result: the offset request round-trips through a real table write, then gets silently undone by the auto-correction that's supposed to handle metering drift, not a request that never took effect in the first place.

Guess at where to look: either register() needs to force/verify an actual seek (not just a table write) before focus_frame/metering runs, or recentered()'s correction needs to distinguish "this doesn't match because the request was a deliberate offset" from "this doesn't match because of normal metering slop" — right now it looks like it can't tell the two apart, so a deliberate offset always loses.

activexray:
Wait how did it apply a frame offset directly in nkscan? I'm not sure I believe claude here, but I'll dig through the trace.

thetalkingdrum:
that part might a hallucination as i tried some headless tests this morning and it is mixing that up. Right now we only tested with negpy

activexray:
Can you point me to the fork on negpy you're working with?

thetalkingdrum:
Claude is retracting part of the statement:) 

To clarify the mix-up: the "raw nkscan.Session" testing was real, but it was against an earlier build (6094418, before your logging change) and done by directly calling nkscan.Session.scan_frame() in a Python shell, bypassing NegPy — not through this PR's code. Everything on 0.11.0/9f855bc was tested through NegPy's own UI only. My earlier comment blurred those two together, sorry for the confusion.

On the fork question: the patch wasn't pushed anywhere when you asked — it was a local, uncommitted change on top of pr-1042-nkscan-improve. I've now pushed it as its own branch so you can check it out directly:

https://github.com/thetalkingdrum/NegPy/tree/debug/nkscan-offset-no-op

(diff against pr-1042-nkscan-improve: https://github.com/thetalkingdrum/NegPy/compare/pr-1042-nkscan-improve...debug/nkscan-offset-no-op)

It's the two files adapted to your Python API changes, plus the init_logging("trace") call and [bleed-debug] logging left in so the trace I posted is reproducible as-is.

thetalkingdrum:
> > Also there seemed to be a mismatch between the preview and the scan. The preview did have some badly framed frames, but the scan came back correctly framed without touching the offset.
> 
> I might need to dig into what negpy means by preview. I don't have a separate "preview" pass from "scan" or "thumbnail". There is one in Nikon Scan, I just haven't wired it up yet. afaik it's just a lower DPI pass.

I think Nikon scan uses the quick thumbnail pass to build a map of all the frames, and might be beneficial when doing full roll scanning. I think negpy's preview (and nkscans thumbnail?) is similar to this in spirit. The function actually called "preview" in Nikon scan is just a lower res pass I believe, not anything like the preview function in Negpy

activexray:
Ok I see what's happening. I need to de-slop some stuff.

marcinz606:
fyi I'm busy with other stuff so I'll probably revisit this during the weekend

thetalkingdrum:
I looked a bit into the preview/scan mismatch thing in case it is helpful: 

Two separate sources of mismatch between the preview and the real scan, not one:

1. Position error. NegPy converts a stage address into a thumbnail pixel by dividing by 41 (optical_dpi / thumbnail_dpi, i.e. 4000/97). But thumbnail::frames() now builds its addresses using the corrected pitch, LinePitch::measured, which for this scanner is 41.555 addresses per line, not 41. Your own test asserts that gap is 76 addresses over one frame's ~137 lines. NegPy's conversion doesn't know about the corrected pitch, so it drifts further off the more lines it's crossed — worse on later frames than earlier ones, always toward the next frame. This is the dominant effect and explains what we actually saw.

2. No edge refinement. Even with the rectangle positioned correctly, the preview just cuts out that raw rectangle as-is. The real scan gets an extra step: after the pass, nkscan looks at the actual pixels and finds the true edges of the picture (frame_columns), then trims to that. The preview has no equivalent — nothing to trim against. Right now this one's small (10-20 address-units out of 5669, under 0.5%, from comparing frame_columns against the coarse rect across several real scans), so it's not what's causing the visible mismatch. But it's a real, separate gap: fixing 1. alone would get the preview correctly positioned, not equivalently precise at the edges.

activexray:
boy I hate llm prose so much. I'm ripping out a lot of slop and fixing more things, although I won't be able to test until I get home later this evening.

thetalkingdrum:
Lol, sorry about that. I know it's annoying.

webmogul1:
> boy I hate llm prose so much. I'm ripping out a lot of slop and fixing more things, although I won't be able to test until I get home later this evening.

you are not alone hahaha

thetalkingdrum:
Seems to be working again now with the latest commit

marcinz606:
all good against `main` nkscan, once 0.11.0 wheels are published this just needs them pinned and it's ready

thetalkingdrum:
Offset works fine now, but I'm still getting mismatch between the preview/thumbnail and the scan. The preview has been perfect for the strips I have scanned so far, with no film base at either side of the frame. But all the scanned frames are shifted somewhat to the right from the preview and shows some black border. 

So everything looks fine in the preview, but the scan is always slightly off. So setting an offset isn't reliable as you are flying blind. At least on my LS-50.

thetalkingdrum:
Thinking about this some more I'm starting to wonder if the SA-21/30 are simply not accurate enough to match the preview with the scan? Nikon scan always overscans and includes film base on both sides of the frame so there is a much bigger margin for error. I think it *might* be a bit more accurate as well, but there are visible mismatches between thumbnail and scan there as well on some frames. So it is not like it manages to position the film perfectly for every frame either.  But the small errors gets absorbed because of the overscan.

oleg-spec:
> Поразмыслив, я начинаю сомневаться, не являются ли SA-21/30 просто недостаточно точными, чтобы сопоставить предварительный просмотр со сканом? Сканирование Nikon всегда включает пересканирование и затрагивает базовую поверхность пленки с обеих сторон кадра, поэтому погрешность гораздо больше. Думаю, оно _может_ быть немного точнее, но и здесь на некоторых кадрах видны несоответствия между миниатюрой и сканом. Так что нельзя сказать, что пленка идеально позиционируется для каждого кадра. Но небольшие ошибки компенсируются за счет пересканирования.

I use modifyed ls-50 with sa21 ( full roll scanning ) with Silverfast . All my corrections with preview i see after scanning in scan folder, they are good

PR #1049, opened 2026-09-04
fix: add optional inter-exposure delay for trichrome capture

matthoosier:
Some older USB/PTP bodies, including the Canon EOS 450D, need a short pause between successive capture commands. A trichrome scan fires red, green and blue back-to-back, and the camera may still be clearing the previous transfer when the next command arrives.

Keep the default at zero for normal cameras, while preserving the user-controlled inter-exposure delay for affected bodies.


matthoosier:
@marcinz606 : this is a small tweak to allow using older camera bodies that choke on the rapid fire trichrome scanning and the calibration to find power levels for the bands that yield equal peaks through the film base.

Do you think this would be okay to merge?

PR #1051, opened 2026-09-05
Offer a writable data folder when Windows blocks startup

hsnilsson:
## Problem

NegPy can exit before showing its main window when Windows Controlled Folder Access blocks writes to Documents/NegPy. Directory existence does not prove write access, and application paths must be selected before configuration is imported.

## Changes

- Check data-folder write access before importing the main application on Windows.
- When the default folder is not writable, show a native dialog that suggests `%LOCALAPPDATA%\NegPy\data`. The user can click the displayed path to open a folder picker and change it, then choose **Use This Folder**, or cancel.
- Explain that no files are copied or moved: a new folder starts with new settings and no saved edits, while an existing NegPy data folder uses its own data. Original data stays in place.
- Provide help for allowing a trusted NegPy executable through Windows Controlled Folder Access, without disabling protection or adding an antivirus exclusion. The user can cancel and restart after restoring access to the original folder.
- Validate the selected folder and atomically save its location under Local AppData. Reuse it on subsequent starts, including when Documents becomes writable again. A failed selection lets the user choose again; an invalid or inaccessible saved location is reported rather than silently replaced.
- Honor `NEGPY_USER_DIR` ahead of the saved location. An inaccessible explicit override gets a visible error.
- Use native controls before Qt initialization so the selected folder's saved UI scale can still be applied before QApplication is created.
- Add regression tests and user/developer documentation.

There is no automatic data migration or SQLite snapshot/backup code in this diff. There are also no Numba-cache changes, orphan-widget fixes, or Defender-setting changes. Previously saved relative location records remain readable so existing selections do not lose access to their data.

## Validation

- Full Windows suite: **5,440 passed**, 19 skipped, 14 deselected; 57 subtests passed.
- Focused startup, folder-selection and native-dialog tests: **41 passed**. Cases include cancellation, custom paths, failed selection and retry, concurrent publication, invalid/missing saved locations, environment precedence, unchanged source/target data, hyperlink handling and startup import ordering.
- Ruff format/lint and the repository-configured type check passed.
- Native Windows UI smoke test: clicked the path, opened the actual shell folder picker, selected a custom folder containing spaces, Å and &, verified the updated display, and accepted that path. Inspected a rendered capture of the dialog.
- Built a Windows PyInstaller executable and verified all **309 bundled NegPy modules plus the entrypoint** against the worktree source.
- Tested that executable against the real protected Documents folder, with the normal Windows user profile, no `NEGPY_USER_DIR` override, and an isolated Local AppData destination. Defender event 1123 confirmed the executable's write was blocked. No Defender settings were changed.
- Cancel exited without creating or selecting the proposed data folder. Accept reached the normal main window and GPU initialization. Restart reused the selected folder without prompting. Both successful runs closed cleanly after closing their main/tutorial windows. Original database/settings/preset files checked by the test remained byte-identical.

The native integration harness and packaged executable are local verification artifacts, not application code in this PR. Choosing another folder does not transfer old edits: users who need their original data must restore access to that folder or transfer the data separately.


hsnilsson:
Hey guys!

I haven't been able to start NegPy since May or something (the process dies after one second on every version except 42/48 or so). 
I haven't seen anyone else report it, but I asked Codex to fix for me locally and I ended up with a PR since it found some actual issues. 

This is of course a sloppy hello, but just give a thumbs up (or down) to let me know if you're interested in taking it in. If positive then I'll review it properly as well as take it an extra round with the new Astra model. I'm currenlty on Windows though, so for testing on other OS:es I would need some help.

marcinz606:
thumbsup for windows fix contributions. I don't want to get anywhere near that so all help is very welcome :)

hsnilsson:
Great! I've been the sole-windows-guy before ;)

PR #1055, opened 2026-09-06
Fix transient small windows during UI startup

hsnilsson:
## Problem

During UI construction, several controls call setVisible(True) before Qt assigns them a parent. They can briefly appear as small standalone windows at startup.

## Changes

Attach the canvas toolbar row to its container before rebuilding it, add the Scanlight setup hint to its layout before applying visibility, and create expanded section content with its section as parent.

Add an application event-filter fixture that catches transient top-level Show events during construction, plus regression tests for all three components. The Scanlight test covers both available and missing gphoto2.

## Validation

- Against unmodified main, the regression tests detect 19 toolbar controls, the missing-gphoto2 hint, and the expanded section frame as unowned windows.
- With the fix, all 152 tests for these components pass.
- Ruff formatting/lint and the repository-configured ty check pass.
- Full Windows suite on commit 347cef7: 5,371 passed, 26 skipped, 14 deselected, and 69 subtests passed; exit code 0. Coverage: 78%. Run with Vulkan, an isolated NEGPY_USER_DIR, and the repository's default test selection (not slow). The complete run finished in 430 seconds; test_gamut_readout.py and subsequent GPU tests passed in the full suite.

This draft is based on upstream main at 2bee1d0. UI regression tests use Qt's offscreen platform; a packaged executable was not rebuilt for this change.


hsnilsson:
Resolves this bug:
https://github.com/user-attachments/assets/144c35a0-1686-48a2-8bf2-e2df72ed6213

The full-suite validation must be run before this is validated. I did it in [#1051](https://github.com/marcinz606/NegPy/pull/1051) but decided to split these as this is a different bug that surfaced from that fix.
- this is now completed ✔️

PR #1056, opened 2026-09-06
fix(plustek): stop compressing multi-exposure scan progress into a false plateau

TobbyTravel:
## Summary
- The GL128 multi-exposure scan progress bar undershot reality: at the short/long exposure pass boundary (a real, correctly-reported 50% pass-fraction milestone from pyopticfilm) it displayed ~46% instead, then briefly jumped to 83% before dropping back down once the long pass started streaming bytes.
- Root cause is in `_make_scan_progress()` (`negpy/infrastructure/scanners/plustek_backend.py`), introduced in #1007: it scaled the whole `[0, 1]` pass fraction by an arbitrary `0.72` and reported a hardcoded `0.83` for the "Preparing long exposure" label, which is discontinuous with the surrounding `0.10 + 0.72 * frac` values.
- Fixed by mapping the pass fraction linearly across the full `0.10-0.85` scanning span (`0.10 + 0.75 * frac`), so the plateau value is continuous with both neighboring passes and the bar never jumps backward. The `0.85-1.0` tail stays reserved for the host-side merge/makeup step, which has no incremental progress of its own.
- pyopticfilm's own progress fraction (`(idx + p) / n_pass` in `session_gl128.py`) was verified correct; this is purely a NegPy-side display/mapping fix.

## Test plan
- [x] `uv run pytest tests/scanners/test_plustek_backend.py -q` (31 passed, includes a new regression test asserting the reported fraction never decreases across the short/long pass boundary — fails against the old formula, passes against the fix)
- [x] `uv run pytest tests/ -q` (5444 passed, 22 skipped; one pre-existing unrelated failure in `test_crosstalk_profiles.py`, unaffected by this change)
- [x] `uv run ruff check` / `uv run ruff format --check` on the changed files
- [x] `uv run ty check` on the changed file

### DIFF
--- a/negpy/infrastructure/scanners/plustek_backend.py
+++ b/negpy/infrastructure/scanners/plustek_backend.py
@@ -119,13 +119,16 @@ def _make_scan_progress(
         if frac >= 1.0 - 1e-9:
             _safe_progress(progress, 0.85, "Merging exposures")
             return
-        if frac > plateau + 1e-6:
-            state["long_started"] = True
-            _safe_progress(progress, 0.10 + 0.72 * frac, "Scanning")
-        elif abs(frac - plateau) < 1e-6 and not state["long_started"]:
-            _safe_progress(progress, 0.83, "Preparing long exposure")
+        # 0.10-0.85 mirrors the acquisition span used below; the leftover 0.15 is
+        # reserved for the host-side merge/makeup work after the last pass lands,
+        # which reports no incremental progress of its own.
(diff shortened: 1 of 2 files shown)


TobbyTravel:
Still inaccurate, and I don't think it's worth the calculation and maintenance cost to fix this. Future changes around multi-pass will likely invalidate this again anyway, so I think it's better to leave it as-is for now.

Issue #1057, opened 2026-09-06
[Feature] From metadata lens correction for embedded distortion and lateral CA data


hsnilsson:
## Problem

When camera-scanning film, I would like NegPy to use supported lens-correction data embedded in the source file instead of requiring manual adjustment. The current Geometry UI exposes a manual `Distortion Correction` slider backed by `distortion_k1`.

## Proposed behavior

Add a **From metadata** button/toggle next to Distortion Correction (or in a Lens Corrections section):

- Make it clickable only when the current source contains supported, usable embedded correction data. Otherwise keep it disabled with a reason, such as no embedded correction data or unsupported metadata format.
- Clicking it explicitly enables the available correction; detecting metadata should not silently apply it. Show which corrections are available and active.
- Distinguish **geometric distortion** (barrel/pincushion) from **lateral chromatic aberration** (RGB channel misalignment/color fringes). If usable data exists for only one, enable only that correction. CA-only metadata must not enable geometric distortion correction. Separate toggles under a shared Lens Corrections section would make this clear.
- A lens name, focal length, or a camera correction on/off flag alone is not an embedded correction model. Supported DNG correction opcodes or supported manufacturer correction coefficients could qualify, subject to format support and validation.
- Use the scanning camera/lens metadata from the source, not archival metadata describing the original film camera/lens. This cannot generally remove aberrations already recorded on the film, and is not a general fix for longitudinal CA.
- Preserve manual correction as a fallback and make its interaction with metadata mode explicit to avoid accidentally stacking the same correction. Avoid reapplying corrections already baked into decoded pixels where this can be determined.
- Save the selected mode with the edit and keep preview/export consistent. In a batch or when copying settings, resolve availability and coefficients separately for each source rather than copying another image's lens coefficients.

## Scope / relationship to existing request

Related to #1033, which proposes Lensfun database support. This request specifically concerns **embedded correction data and a metadata-aware UI**. Looking up an external Lensfun profile using EXIF lens identification is a useful complementary path, but should be identified as profile-based correction rather than embedded correction data. These could share an implementation/UI if preferred.

An initial version could support a documented subset of metadata formats, keeping unsupported cases disabled. No sample RAW file has been validated for this request, so it does not assume every camera embeds usable coefficients.

## Acceptance examples

- Supported distortion data: From metadata is available and produces matching preview/export geometry.
- Supported lateral-CA data only: CA correction is available; distortion remains unavailable.
[...]

Issue #1033, opened 2026-08-31
[Feature] use lensfun database for quick (scanning) lens corrections


marcinz606:
There is conventient python wrapper https://pypi.org/project/lensfunpy/ that should make implementation easy

PR #1058, opened 2026-09-06
Add From metadata lens correction for Sony ARW and DNG

hsnilsson:
Camera scans can contain lens-correction coefficients, but the Geometry panel only offers a manual radial coefficient. This adds **From metadata** beside Distortion Correction and shows the source and available distortion/lateral CA correction. Missing, identity, malformed, or unsupported data leave correction unavailable.

Only the mode is saved with the edit. Each source supplies its own coefficients, so copied settings and mixed batches cannot apply another file's lens profile. Enabling the mode replaces manual distortion; disabling it restores the manual slider.

### Implementation

- `LensWarp` defines `has_distortion`, `has_ca`, and `remap(...)`. Frozen `RectilinearWarp` and `SonyWarp` classes own their coordinate calculations. Metadata and rendering use the protocol without vendor-specific branches.
- One extension-to-reader table selects a file reader. Sony and DNG share bounded TIFF traversal; other containers can supply their own reader. Adding a model requires a warp class, a reader, and one registration.
- Preview and export share the same inverse warp, after flat-field and before sensor unmix, user geometry, and crop. Row blocks bound map memory. Flat-field is not applied again downstream.
- EXIF orientation, DNG ActiveArea, and decoder/DefaultCrop offsets are respected, including the existing DNG 1.7 JPEG XL fallback.
- The mode and flat-field content are part of source/cache identity. Positive Source remains independent in the decode paths and preview cache.
- Existing sidebar controls and shortcut registration are reused. The in-app guide, pipeline documentation, and architecture notes are updated.

### Supported scope

Sony ARW distortion and red/blue CA tables, plus DNG **WarpRectilinear** (opcode 1, one or three planes). DNG 1.7 is supported when it carries this correction model. Lens identification and camera on/off flags alone do not enable correction; inherited Sony tags in converted DNG files are not applied.

WarpRectilinear2, fisheye, other/mixed opcode lists, composites, and RGB+IR sources remain unsupported. Choose the correction before crop/retouch. Edge sampling replicates boundary pixels. No Lensfun lookup is added.

### Validation

- Full suite: **5,496 passed, 1 failed, 26 skipped, 14 deselected** on Windows / Python 3.13.
- The failure was `test_saving_again_keeps_the_profile_in_its_own_file`: a same-mtime/size TOML cache collision in code unchanged from `upstream/main`. It was reproduced separately; all **17 crosstalk profile tests passed on rerun**.
- **61 metadata tests** cover validation, TIFF byte orders, per-channel CA, all EXIF orientations, per-file coefficients, preview/export parity, and independent Positive Source/lens cache entries.
- A third structural warp and a non-TIFF test reader exercise registration, capability reporting, caching, orientation, channel mapping, and row blocks without changing shared code.
- **32 image comparisons are pixel-identical to the implementation before the protocol refactor**, covering Sony, DNG, sequential warps, crop offsets, and all EXIF orientations.
- Ruff formatting/lint and the repository's ty checks pass; the protocol and warp modules also pass ty without diagnostic exclusions. Makefile recipes were run through `uv` because GNU Make is unavailable on this host.

The user's Sony A7R V files have not been tested directly.

Implements the initial supported-format subset of #1057. Related to #1033 (external Lensfun profiles).


marcinz606:
nice, I had quick scrollthrough and I have one suggestion:

right now adding a new vendor (eg Fuji or Nikon) means touching few places. Consider giving RectilinearWarp/SonyWarp a common tiny protocol (has_distortion, has_ca, a remap(...) method) so a future CanonWarp is one new class + one new reader function, not three edited call sites.

hsnilsson:
Hi, my agent pushed this without my permission and before I had time to actually work on it. Sorry for that, and thanks for great feedback - definitely makes sense to have a protocol. I hope to get some time tomorrow or at least within a few days to finish this.

hsnilsson:
Status update: This PR is my goal here but have had troubles running NegPy since I got it working on my machine. Continued memory issues with my 244 MP DNG JXL images. Working on a thumbnail PR but finding other memory issues still. Of course this is high demanding images but if NegPy is going to be useful for me I need those to work.

hsnilsson (commented):
This is ready for further review.

PR #1060, opened 2026-09-07
feat(exposure): Positive Source for an already-finished Transparency

PabloRuizCuevas:
## What

Adds a **Positive Source** toggle (Calibration → Capture, Transparency + Normalize off only) for loading an already-finished positive image — a scanned print, an export from other software — as a Slide, instead of a raw scanner/camera capture.

## Why

The as-captured Transparency transfer is built entirely around a raw capture: it decodes the source as literal linear data, adds a fixed **+0.7 EV baseline lift**, and runs a **filmic highlight roll-off** (Narkowicz's ACES RRT fit) — all correct for an underexposed sensor capture, all wrong for a source that's already correctly exposed and graded. Feeding an already-positive TIFF through this path renders it badly blown out and flattened, with no existing control to fix it.

## What Positive Source does, when on

- The loader decodes the source's own embedded ICC profile (sRGB or Adobe RGB) instead of assuming linear data. An **untagged** file gets the same sRGB assumption an untagged 8-bit file already gets — `positive_source` has already resolved the ambiguity a bare untagged 16-bit file otherwise carries (scanner-linear vs. finished positive).
- The baseline lift and filmic roll-off are skipped, on **both** CPU and GPU, so the Print sliders (Density, Grade, Toe/Shoulder, etc.) shape the capture directly instead of a re-rendered version of it.
- The GPU shader carries the flag through the transfer uniform's previously-unused `zone_taper.y` lane rather than adding a new uniform field.

## Verified

Live, against real files (not just unit tests) — a positive TIFF that rendered blown out (mean 0.82, 43% of pixels clipped past 0.9) now renders within ~4% of its own source values (mean 0.55, <1% clipped), with **no manual pre-tagging** required — same result whether the file is untagged or already carries an sRGB/Adobe RGB profile.

All CPU/GPU parity tests pass (`TestGpuTransferParity`), plus new coverage:
- `tests/test_effective_linear_raw.py` — `positive_source` exempts the forced-linear decode, only on the transfer path
- `tests/test_tiff_loader_encoding.py` — sRGB/Adobe RGB decode, and the untagged-assumes-sRGB fallback
- `tests/test_transparency_transfer.py` — CPU identity + GPU parity for the baseline/ODT skip
- `tests/test_sensor_sidebar.py` — UI gating (live only on the as-captured transfer)

Full suite: 5435 passed, one pre-existing unrelated failure (a macOS Qt tooltip-rendering quirk, present identically on `main` before this change).

## Docs

`docs/USER_GUIDE.md` (§4.1) and `docs/PIPELINE.md` (Transparency transfer section) updated.


thetalkingdrum:
Great! I was thinking of looking into this, but it got put on the back burner as I don't really shoot slides.

PR #1065, opened 2026-09-07
feat(assets): per-file half-frame split, auto-detect or manual override

PabloRuizCuevas:
## Problem

Half Frame's crop/split/gutter has been one shared profile for the whole roll since it shipped: set once in the rectangle editor, then applied identically to every scan regardless of that scan's actual gutter position. A roll with a bit of irregular frame spacing gets some frames' splits wrong, with no way to fix them short of abandoning the shared profile entirely.

The per-file pieces already existed and were simply unused: each expanded half already carries its own `split_x`/`crop_rect`/`gutter_thickness`, and `detect_split_x_for_file` already finds a scan's real gutter — it just stops running the moment a profile is saved (`_expand_half_frames` only calls it when `profile is None`).

## Fix — two independent, backward-compatible additions

Existing behavior is unchanged when neither is used (no profile → still auto-detects everything, same as today; a profile with the new option left off → same fixed split for every scan, same as today).

**1. Auto-detect split per frame** — a checkbox in the roll-wide rectangle editor. On, each scan's own gutter wins over the fixed `split_x`; `crop_rect`/`gutter_thickness` still come from the shared profile (they describe the scanner rig, not the film). A frame whose gutter can't be found (`detect_split_x`'s own 0.5 "nothing found" sentinel) falls back to the profile's tuned value rather than a blind re-center.

**2. Per-frame override** — right-click a half-frame asset → **"Adjust split for this frame…"** opens the same editor scoped to just that file, saved separately from the roll-wide profile (the roll-wide auto-detect checkbox is hidden — meaningless for a single fixed frame). **"Reset split to roll default"** appears once an override exists, to drop back to the shared setting. Wins over both the profile and auto-detection, for the odd frame that still gets it wrong even with #1 on.

Resolution order in `_expand_half_frames`: **per-file override > per-file auto-detect (profile's `auto_split` on, or no profile saved yet) > the profile's fixed `split_x`.**

## Tests

- `tests/test_half_frame.py` — new cases for auto_split on/off and override precedence, alongside the existing `_expand_half_frames` coverage
- `tests/test_half_frame_dialog.py` — new file: the auto-split checkbox shown/hidden/checked correctly, dialog title
- `tests/test_controller.py` — profile/override storage round-trips, `open_half_frame_dialog`'s per-frame vs roll-wide branching
- `tests/test_file_browser_widget.py` — context menu shows the new actions only when relevant

`docs/USER_GUIDE.md` updated (Half Frame section).

Full suite: 5457 passed (up from 5438), one pre-existing unrelated failure (macOS Qt tooltip-rendering quirk, present identically on `main`).


negpy/desktop/view/sidebar/files.py
```diff
@@ -434,9 +435,28 @@ def _init_ui(self) -> None:
 
         self.half_frame_adjust_btn = QToolButton()
         self.half_frame_adjust_btn.setIcon(qta.icon("mdi.tune-variant", color=THEME.text_primary))
-        self.half_frame_adjust_btn.setToolTip("Adjust Half Frame split — reposition the crop rectangle and split line for the current scan")
+        self.half_frame_adjust_btn.setToolTip(
+            "Adjust Half Frame split — reposition the crop rectangle and split line; its own Apply ▾ picks what to apply it to"
+        )
         self.half_frame_adjust_btn.clicked.connect(self._on_half_frame_adjust)
 
+        self.half_frame_auto_all_btn = QToolButton()
```
marcinz606, line 443: I don't like so many half-frame specific buttons put in here, maybe instead have one half-frame specific button that brings up the menu from which you can select all the related options?

PabloRuizCuevas: Fair — collapsed Adjust split, Auto-detect all splits and Unsplit diptych into one button with a menu (`dfbe763`), so it's back to one half-frame-specific icon plus the mode toggle.

PabloRuizCuevas:
Screenshot 2569-09-12 at 03 16 51

negpy/desktop/view/sidebar/files.py
PabloRuizCuevas: ok changed

PR #1075, opened 2026-09-09
fix(export): stop the filename field from stealing the cursor while typing

### LINKED ISSUE #1071: [Bug]: export filename field constantly moves cursor to end of line
### NegPy version / build

0.58.0

### OS

Windows 11

### Scanner

N/A (not a scan-related issue)

### Scanner model / details

_No response_

### Light source

N/A

### Correction mode

N/A

_No response_

_No response_

### What happened

In the export tab, the Filename field (where you can put placeholders and whatnot) keeps moving the cursor to the end of the line roughly once per second, even while typing, which makes it nearly impossible to actually type in that field. I had to type out my format in notepad and paste it in.

### Source file

N/A

### Screenshots

This field is the one I'm talking about: 

Image

TobbyTravel:
## Summary
- Fixes #1071 — the Export tab's Filename field kept resetting the caret to the end of the line roughly once a second, making it nearly impossible to type into.
- Root cause: `ExportSettingsForm.load()` runs on every debounced AppState resync (`config_updated` -> 150ms timer in `right_panel.py`). Typing in the Filename field routes through the export sidebar's own 500ms persist debounce (`update_timer` in `export.py`), which writes the value back to AppState and fires `config_updated`, which calls `load()` again. `QLineEdit.setText()` unconditionally moves the caret to the end, even when the text is unchanged, so any pause of ~650ms between keystrokes yanks the cursor to the end.
- Fix: skip `setText()` on the destination text fields (`filename_edit`, `subfolder_edit`, `abspath_edit`) while they have focus — the field's own `textChanged` handler already keeps AppState current, so there's nothing to restore.

## Test plan
- [x] Added `test_load_does_not_reset_cursor_in_focused_filename_field`, which reproduces the bug by focusing the field, positioning the caret mid-string, then calling `load()` again as the debounced resync would — fails before the fix (caret jumps to the end), passes after.
- [x] `uv run pytest tests/test_export_settings_form.py -v` — 27 passed
- [x] `uv run pytest tests/ -q` — full suite passes (5437 passed, 26 skipped, 14 deselected)
- [x] `make format` / `ruff check` — clean
- [x] `ty check` on the changed file — no new diagnostics (one pre-existing unrelated warning at line 536)

### DIFF
--- a/negpy/desktop/view/sidebar/export.py
+++ b/negpy/desktop/view/sidebar/export.py
@@ -1597,7 +1597,10 @@ class ExportSidebar(BaseSidebar):
             self._cs_background_color = layout.background_color
             self._cs_label_color = layout.label_color
             self._update_cs_colors_btn_tooltip()
-            self.cs_output_path_edit.setText(conf.contact_sheet_output_path)
+            # setText() unconditionally moves the caret to the end; skip the refresh while the
+            # user is actively editing the field, same as ExportSettingsForm._set_text_preserving_edit.
+            if not self.cs_output_path_edit.hasFocus():
+                self.cs_output_path_edit.setText(conf.contact_sheet_output_path)
             self.sidecars_enabled_btn.setChecked(conf.export_sidecars_enabled)
             self.printing_notes_preview_btn.setChecked(self.state.printing_notes)
             self._refresh_contact_sheet_templates()
--- a/negpy/desktop/view/widgets/export_settings_form.py
+++ b/negpy/desktop/view/widgets/export_settings_form.py
@@ -676,6 +676,16 @@ class ExportSettingsForm(QWidget):
 
     # --- Load / read ---------------------------------------------------------
 
+    @staticmethod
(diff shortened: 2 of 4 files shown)


TobbyTravel:
Code review caught that \`cs_output_path_edit\` (Contact Sheet → Output path) goes through the identical \`config_updated\` → 150ms \`_sync_debounce\` → \`sync_ui()\` round-trip as the Filename field, and was still calling \`setText()\` unconditionally — same bug, same field class. Pushed a follow-up commit applying the same focus guard there, plus a regression test (\`test_sync_ui_does_not_reset_cursor_in_focused_contact_sheet_path\`) that fails without the fix and passes with it.

I checked the rest of \`export.py\`/\`export_settings_form.py\` for other unconditional \`setText()\` calls on user-typed fields — the remaining ones (\`cs_output_path_edit\` from the folder-picker, \`abspath_edit\` from the folder-picker, and various label/button text) are all direct responses to a user action, not periodic resyncs, so they're fine as-is.

PR #1090, opened 2026-09-13
feat(process): highlight reconstruction for Transparency (#1072)

### LINKED ISSUE #1072: [Feature]: Highlight reconstruction for RAW files
### NegPy version / build

0.58

### OS

macOS 26

### Scanner

Digital camera (macro/copy setup)

### Scanner model / details

Sony A7 + Nikkor 55mm f/3.5

### Light source

CIneStill CS Lite

### Correction mode

Off

_No response_

_No response_

### What happened

When clipping occurs in RAW files, clipped highlights may appear as pink, because of the Bayer sensor and the way white balance is handled.
Although this is an edge case and of course it is preferable to avoid clipping, it is sometimes not really avoidable, like when scanning slides with very thin parts (like the sun).

It would be great if clipped highlights would be reconstructed in this case. Darktable uses the "inpaint opposed" method, which makes highlights clipt to true white instead of pink:
https://docs.darktable.org/usermanual/5.6/en/module-reference/processing-modules/highlight-reconstruction/

I'll attach an example scan (which is not a masterpiece of photography, but shows the case) and its NegPy interpretation.

### Source file

[2017-04-09_Dia_Film2017015_012.ARW.zip](https://github.com/user-attachments/files/32026185/2017-04-09_Dia_Film2017015_012.ARW.zip)

Image

### Screenshots

_No response_

thetalkingdrum:
Adds libraw's highlight reconstruction (HighlightMode) to the Transparency
(E-6) transfer path, exposed as a Highlight Recovery dropdown (Off/Blend/
Reconstruct) in the Process panel, visible on Transparency and greyed out
on a source with no camera matrix (a scanner TIFF, JPEG, or other already-
rendered file, which has no sensor data left to recover from).

- New `ProcessConfig.highlight_reconstruction` field, gated everywhere through
  `effective_highlight_reconstruction` — forced to Off outside E-6 or under
  narrowband, so a stale/copied value can never reconstruct a negative's
  clipped base.
- The as-captured transfer path decodes neutral by default; an active
  reconstruction bakes the real white balance into the decode instead (real
  clipping is invisible to libraw on a neutral decode) and skips the
  downstream matrix fold to match.
- Compensates libraw's own highlight-mode exposure scaling (a non-Clip mode
  normalizes against the widest channel multiplier instead of the narrowest,
  darkening the whole decode) so turning reconstruction on doesn't shift
  overall exposure.
- Highlight Reconstruction is dropped on a merged (HDR) bracket entirely — WorkspaceConfig.__post_init__ zeroes it the same way it already drops e6_normalize, and HdrWorker.run's solve zeroes it explicitly too, since it decodes each frame before the merge's HDR config exists to trigger that invariant. A reconstructed pixel no longer reads near the sensor ceiling, so a bracket's own clip detection would otherwise trust inconsistent per-frame guesses as real signal and blend them into artifacts; a bracket already recovers genuine highlights from its shorter, unclipped exposure instead. Documented in USER_GUIDE.md and PIPELINE.md.
- UI: a single Highlight Recovery dropdown row, matching the pattern
  Demosaic already uses for this kind of occasional, discrete choice.

Closes #1072.


thetalkingdrum:
@seanharding maybe want to have a look as you know this part of the codebase well?

seanharding:
Sure, I can poke at it a little bit later today.

seanharding:
This seems to create weird color shifts in HDR merges.

Screenshot 2026-09-13 at 12 00 30
Screenshot 2026-09-13 at 12 00 34

seanharding:
I don't really have time to mess with it more today, but this is Claude's assessment, if it's helpful...

**A merged bracket's preview loses the white-balance fold once Blend or Reconstruct is on.**

With reconstruction active on the Transparency transfer path, highlight_reconstruction_bakes_wb is true, so should_fold_camera_wb returns false and both engines skip the camera-matrix fold. That is right for export and for the bracket solve, which bake the reference's white balance into every frame's decode.

The bracket preview does not bake, though. load_linear_preview_hdr forces bake_camera_wb = False and decodes every frame neutral, but the render still derives the fold decision from the config, not from what the decode did. So the preview buffer carries neither the baked white balance nor the fold, and the merged bracket renders with the raw green-to-red cast. The same path feeds decode_asset_preview, so batch autocrop and contact sheets see it too.

The USER_GUIDE and PIPELINE text describe this as "the preview does not reflect Blend or Reconstruct". The actual gap is a full white balance, so the export will not match what the user saw in the preview.

Two possible fixes: pin the bracket preview's frames to the reference's white balance the way export already does, or make the fold follow what the decode actually did (a flag on the loaded buffer) instead of re-deriving it from the config.

Repro: E-6, Normalize off, a merged bracket, set Highlight Recovery to Blend, compare the preview to the export.

seanharding:
I noticed another problem that looks like it was introduced by #953. Submitted #1092 to fix it.

thetalkingdrum:
Thanks for testing. I don't have the files available to do the bracket test. I went with Claude's first fix. I was thinking of gating it off as well as it is kind of a corner case, but decided to let it stay. Could you test it again to see if the fix works when you get the chance?

thetalkingdrum:
> I noticed another problem that looks like it was introduced by #953. Submitted #1092 to fix it.

Nice catch.

seanharding:
Two issues I see right now.

First, "peek negative" goes wonky with highlight recovery on. (Not totally sure that "peek negative" even makes sense on a transparency, so maybe we should disable it, but if it's going to be there it should work. And I'm not sure if this exposes any other underlying problems).

Screenshot 2026-09-14 at 12 39 31

Second, the text under the highlight recovery dropdown says that it's not applied to a merged bracket, but it seems like it is. In this example, it creates artifacts in the sky. First screenshot with it disabled, second screenshot with it enabled.

Screenshot 2026-09-14 at 12 48 51

Screenshot 2026-09-14 at 12 49 18

I don't think either of these are massive problems. Like I said, maybe peek negative doesn't even make sense in this context. It's definitely not what I had in mind originally for that feature. And if highlight recovery causes merges to look bad, people can just... not enable it. But we should make sure the UI text is accurate at least, and if it can never work properly on a merged bracket we should probably ensure that it is never applied to one.

thetalkingdrum:
Thanks again. Yeah, I'm not sure if peek negative makes any sense on positives, but I fixed the bug anyway. And I gated off for merged brackets now. It is anyway not needed there I think. The main feature seems to work quite well for what it was intended to do, so maybe it's ready to go now unless you see any other problems?

seanharding:
Good to go from my perspective.

PR #1104, opened 2026-09-15
fix(preview): bound GPU, thumbnail, and prefetch memory

hsnilsson:
AI-assisted comment (model: GPT-5).

## Summary

- Decode difficult thumbnails and LinearRaw previews with bounded background work so the UI stays responsive and duplicate foreground/background decoding is avoided.
- Tile GPU preview work, account for auxiliary cache buffers, and release render resources deterministically.
- Keep neighbor prefetch automatic, but run it only when preview-cache capacity and available system memory provide enough headroom. Serialize background decoding and cancel stale work during navigation.
- Bound IR downsampling, preserve IR defects across streamed RGBI segment boundaries, and keep half-frame native dimensions correct when previews use reduced RAW decodes.
- Update the user guide, pipeline documentation, and regression coverage.

## Verification

- Ruff formatting and lint passed.
- Ty type checking passed.
- 577 focused controller, loader, cache, GPU, IR, half-frame, and thumbnail tests passed.
- Full suite with UTF-8 mode: 5,730 passed, 26 skipped, and 14 deselected.
- One unchanged upstream Windows test remains failing: `tests/test_crosstalk_profiles.py::test_saving_again_keeps_the_profile_in_its_own_file`. The TOML cache can reuse stale data when two same-length writes receive the same file modification timestamp. This branch does not change that test or the crosstalk/cache implementation.
- Large segmented LinearRaw DNG files were exercised through the bounded foreground and prefetch paths. Cache/RAM rejection happened before decoding, prefetch work stayed serialized, cancellation completed cleanly, and no OOM, GPU-validation, device-lost, or Qt thread errors were found.
- Explicit full-resolution HQ decoding remains limited by LibRaw's native full-frame CPU allocation and cannot be cancelled inside the native call.


hsnilsson:
I'll split this into 2-3 smaller PRs. Don't waste time on reviewing just yet.
This is on the path to resolving my major blockers when handling large pixelshift images. I would normally speak in person with some other team member before pushing something like this ;)

hsnilsson:
The other PRs are now pushed and should be merged in order https://github.com/marcinz606/NegPy/pull/1107 -> https://github.com/marcinz606/NegPy/pull/1119 -> https://github.com/marcinz606/NegPy/pull/1121 -> https://github.com/marcinz606/NegPy/pull/1122

PR #1105, opened 2026-09-15
Roll workflow, Gear/Metadata reorganization, half-frame detection

PabloRuizCuevas:
## Summary

A session's worth of work across roll handling, gear/metadata organization, half-frame detection, geometry and film-strip UI.

### Roll workflow
- Rolls as a first-class library concept (folder rolls + hand-picked/search-built virtual rolls), with a Roll Settings dialog for tagging gear/process/capture metadata across a frame, a selection or a whole roll.
- Export destination now defaults to the roll's own folder + an `export` subfolder (falling back to `negpy/<roll name>/export` with a warning for a virtual roll with no single folder). A one-time migration drops a still-untouched sticky destination from before this default existed, without touching anyone's deliberately chosen path.
- Roll Analysis redesigned: the separate "BATCH"/"ROLL" sections merge into one, the roll picker is now searchable with a **Current Roll** default, and a single primary **Apply** button replaces the magnifying-glass "Batch Analysis" button that read as a search control. Now sits before Normalization in the Setup tab, since it feeds it a baseline rather than the other way around.
- Roll Analysis pools each frame's own black/white points into an interquartile-trimmed baseline, then reports which frames were trimmed out of that pool as outliers — the roll average is, by construction, the worst match for exactly those frames, and previously nothing surfaced them. **Lock Bounds** now genuinely exempts a frame from Roll Analysis (it used to be silently re-enabled on every re-run), and its control moves onto the roll picker's own row, next to what it actually exempts a frame from. The picker now sits below the Analysis Buffer/Region controls that feed it, with a **Reanalyze** button beside it — the same Batch Analysis the Library's "Analyze Roll…" already offers, as a second, more discoverable entry point.
- Camera, film stock, ISO and capture date can now be inferred from a roll's folder name (`04_om1_Fuji400_Japon` → Olympus OM-1 + Fuji 400; `06_scala_50_vietnam` → ISO 50), gear matched only against your own library, ISO/date read directly from any plausible number or date in the name. Previously only camera/film stock were inferred, and only reachable indirectly through Roll Settings; now also a button on the Metadata tab.
- Roll-wide defaults for Calibration, Demosaic, Normalization and Film Mode/Positive: set once on the roll, every member frame follows, with a per-frame **Lock** that breaks one frame away from the roll's value and releases again on its own once that frame's value matches the roll's. A split button (All Roll / Current Frame / Selected Frames) picks what the next write targets, mirroring Export's own scope control; roll-level vs. frame-level state is now visually unmistakable in the Roll tab.
- An opt-in **per-roll edit fork**: a photo shared by more than one roll normally has one edit everywhere it's opened — "Edit Independently in This Roll" (right-click a frame in 2+ rolls) gives it an independent edit for one roll alone; "Use the Shared Edit Again" reverts it. Reuses the half-frame scan's hash-suffix identity trick, so undo history, thumbnails and the render cache pick it up for free.
- **Reset Roll to Defaults**, on the Film Strip section's ⋮ menu next to New Roll: resets every frame in the roll the same way Reset Settings already resets one, with a confirm dialog naming the frame count.
- Renaming a folder roll can now optionally rename the folder on disk too (opt-in, asked fresh every time, with a cloud-sync warning) — previously a rename only ever touched the library's own record.

### Exposure
- Auto Density and Auto Grade now work on a Positive frame (a scanned print, an export from other software, a scanner-positivized negative) too, instead of staying inert the way they used to on the whole transfer path, restated on the transfer curve instead of the paper's toe/shoulder one — but they default off rather than on for it, the same per-shot-neutral convention White/Black Point already follow, since a Positive frame's exposure is an already-settled decision, not a bracket to preserve. A one-time migration applies that default to every frame already marked Positive, and a scoped Reset now restores it correctly instead of the flat always-on default a negative uses.
- White Point/Black Point now work on the transparency transfer path too, instead of being skipped outright: the offsets are additive and zero at their neutral default, so they deviate the fixed transfer window the same way they already deviate a measured one.

### Gear & Metadata
- Gear tab defaults to the gear you've declared as your own, with the full shipped catalog reachable via a Catalog toggle. Split into two subtabs, **Items** (cameras, lenses, film stocks, processes, scan setups) and **Presets** (saved metadata field sets) — these were previously commingled in one category list despite sharing almost nothing structurally.
- The Metadata tab's own gear combos (camera/lens/film stock/process/scan setup) got the same personal-only-by-default treatment, with an **Other…** entry to reach the full catalog or add a custom one.
- Fixed Add Custom leaving a literal "New Camera"/"New Lens" placeholder in the real `make`/`model` fields, which rode straight into exported EXIF; the placeholder now lives only in the UI-only `display_name`.
- **Protect Original Metadata** moved from the Metadata tab to the Export tab, next to Sync To Batch — both are export-time behaviors over the Metadata tab's fields, not metadata content themselves.
- Metadata Preview and Metadata Presets cards are now always expanded (no collapse chevron) — both are small, glance/quick-action content where collapsing cost more clicks than it saved space; the heavier per-category cards stay collapsible.

### Half-frame
- Auto-detect now finds the outer film crop from each edge inward, not just the center gutter, for tight camera-scan captures with no scanner bed.
- The half-frame toggle auto-detects splits directly when turned on, no dialog.
- Half-frame mode is now remembered per roll rather than globally.

### Geometry
- **Crop by Default**: a toggle in Alignment that crops out the wedge Fine Rotation, Tilt and Swing leave behind, so no edge shows replicated/extrapolated pixels. Applies live whenever no manual or detected crop is set; a drawn or detected crop still takes over from it. While a slider is being adjusted, the canvas briefly darkens the margin it would trim, fading out after it settles, the same as Fine Rotation's own alignment grid.

### Film strip / UI
- "New Roll" moved off the toolbar (where it was forcing a wrap to a second line at narrow widths) into a small always-visible actions menu on the Film Strip section's own header.
- A stale-thumbnail indicator flags a frame whose thumbnail no longer matches its current edit.
- The Negative Stats panel packs two stats per row instead of one.

### Housekeeping
- A sticky-settings master toggle in Preferences.
- Normalize's row now travels correctly through Apply Settings (a pre-existing gap in the settings catalog).
- Force Settings now sweeps stray per-frame locks across the whole roll, and the Roll tab's Apply button grays out when the current scope has nothing to write.

## Testing

Every commit passed `make all` (lint + type + tests) individually; the full combined branch passes the same, plus a full local build + manual launch/smoke check. One pre-existing, unrelated failure throughout (`tests/test_metadata_presets.py::TestPresetNames::test_load_tooltip_follows_a_rebinding`, a macOS key-symbol rendering quirk that reproduces on unmodified `main`). CI (Test and Lint, Preview load metrics) is green.


marcinz606:
@PabloRuizCuevas i would merge it as it is now unless you want to change something before that?

PR #1115, opened 2026-09-15
Film strip: two-column negative stats, stale-thumbnail indicator

PabloRuizCuevas:
## Summary

- **Negative stats pack two per row**: the negative-stats readout uses a two-column layout instead of one long stacked list.
- **Stale-thumbnail indicator**: a bulk apply (Apply Settings, a preset, a roll-wide toggle) writes a non-active file's history without rendering it, so the film strip flags that thumbnail — an amber dot on the cell, a tooltip note — until a real render catches up. Discarded once that frame actually renders, or its thumbnail is dropped on unload.

## Bug fix included

The indicator as originally written never actually showed: the write side (`push_external_history`) flags a frame under `asset_thumbnail_key(asset)` — the same hash-plus-cache-version key used everywhere else in the thumbnail cache, since a trichrome triplet's key differs from its plain hash — but the two read sites (the film strip's dot, its tooltip) checked the bare hash instead, which never matched. Fixed both read sites to key the same way, and added a regression test that exercises the real `push_external_history` → read path end to end (the original test suite had accidentally masked this by seeding its fixture directly with the bare hash, bypassing `asset_thumbnail_key` entirely).

## Testing

`make all` (ruff check/format, full pytest) passes with only one known unrelated failure (`test_load_tooltip_follows_a_rebinding`, a macOS key-symbol rendering difference, pre-existing on `main`).


thetalkingdrum:
Curious how this interacts with #1127. That PR wires `frames_edited_offscreen` into each bulk-write path (Apply Settings, Apply Preset, Batch Analysis, …) one at a time, and #1118 needs the same treatment for its new `reset_roll_settings`. If this indicator flags staleness at `push_external_history` instead, that's a single choke point every bulk writer already goes through — which could catch all of those cases (including future ones) without per-call-site wiring.

Is that the intent here, i.e. should #1127 defer to this instead of adding its own signal? Want to figure out the right order before #1127 gets rebased on top of #1118 and does the wiring by hand.

PR #1123, opened 2026-09-16
fix(files): stop Qt's per-click selection recompute from corrupting shift/ctrl multi-select

thetalkingdrum:
## Summary
- Thumbnail shift-click / Ctrl-click multi-selection in the file browser could silently collapse to one item mid-gesture, on both trackpad and mouse. Root cause: `QAbstractItemView` recomputes its own selection command independently at press, each drag-move, and release, each reading modifiers fresh — a Shift/Ctrl key released a moment before the mouse button let a later stage silently override a correct press-time click. `ThumbnailGridView` now decides a click's mode (plain / Shift-range / Ctrl-toggle) once, from the modifiers at press, and reapplies that decision through the rest of the gesture — a held-button drag still lives-extends the range.
- `AssetListModel`'s persistent-index remap now keys by each row's cached identity instead of its list position, so a sort, a filter, or `refresh()` after a file is added/removed no longer drifts the live selection or the shift-click anchor onto an unrelated file.
- `sync_ui()` no longer lets a stale echo of an in-flight click override a genuinely newer selection command landing in the same debounce window.
- A thumbnail whose decode hasn't finished yet still paints its selection border, so a multi-selection no longer looks incomplete depending on load timing.

## Test plan
- [x] `tests/test_thumbnail_shift_click_selection.py` (20 cases) and `tests/test_filmstrip_placeholder_selection.py` cover the mechanism directly, including ablation-verified regression tests for the modifier race, the anchor-persistence bugs, the debounce race, and the persistent-index remap bug
- [x] Full suite passes (`make test`), lint/format/type clean (`make lint`, `make type`)
- [x] Verified live in the running app on both trackpad and an external mouse


thetalkingdrum:
Merge conflict resolved — kept both the loading glyph/reveal animation and this fix's border-pen/keeper-reject badges in the same block.

That surfaced an interaction worth flagging: the new placeholder `sizeHint()` returns the full cell size instead of a small text-based default, which changes how Qt's autoscroll behaves during a click-and-hold drag. Qt's own `mouseMoveEvent` autoscrolls the viewport before this fix's code re-hit-tests the same screen point, so the hit test lands on a farther row than the user actually dragged to. Fixed by resolving the target row before calling `super().mouseMoveEvent()`/`mouseReleaseEvent()` instead of after, in 6ead1b0.

PR #1107, opened 2026-09-15
perf(thumbnails): stream filmstrip previews in background

hsnilsson:
AI-assisted comment (model: GPT-5).

## Summary

- Let roll discovery finish before thumbnail generation and keep thumbnail work outside the foreground batch workflow.
- Process pending thumbnails one at a time on the thumbnail thread, publish each result immediately, and cancel stale queues when the loaded assets change.
- Keep placeholders visible while background work runs, with a subtle in-cell reveal indicator for the active frame instead of canvas or batch progress.
- Keep rendered thumbnails authoritative and ignore late results for frames that are no longer loaded.
- Remove the obsolete parallel thumbnail batch coordinator.

## Relation to #1104

This is the first focused extraction from [#1104](https://github.com/marcinz606/NegPy/pull/1104). As noted in [the split comment](https://github.com/marcinz606/NegPy/pull/1104#issuecomment-5677987574), the larger draft is being divided into smaller PRs before review.

This PR contains only filmstrip orchestration and presentation. Source decoding is unchanged. Bounded DNG and LinearRaw loading, adaptive GPU prefetch and memory accounting, and the half-frame and IR preview fixes will follow separately.

## Verification

- Ruff formatting and lint passed.
- Ty type checking passed.
- 286 focused thumbnail, controller, and filmstrip tests passed after the final cleanup.
- Full suite with UTF-8 mode and coverage: 5,648 passed, 26 skipped, and 14 deselected.

PR #1124, opened 2026-09-17
Batch rotate and flip across a multi-selection

thetalkingdrum:
## Summary

Rotate and flip (`[`/`]`, the toolbar buttons, their overflow-menu entries) used to only touch the single active frame. Now they apply to the whole multi-selection: select a few thumbnails, rotate, all of them turn — each by its own current orientation, not a copy of the active frame's new one.

- Filmstrip thumbnails update immediately, without opening the frame.
- Undo works per frame, even one you haven't opened yet.
- If the active frame has been explicitly ctrl-click-deselected from a multi-selection, it's left alone (matches `toggle_mark`'s selection-wins rule) instead of turning anyway just because it's on screen.
- Two open copies of the same file (same content hash) only turn once, not twice.

## How it works

- `features/geometry/logic.py`: pure `rotate_geometry_and_analysis` / `flip_geometry_and_analysis`, extracted from the toolbar's inline code.
- `desktop/session.py`: `rotate_selected_frames` / `flip_selected_frames` — for every other selected frame, load its own saved config, transform, record undo history, save.
- `desktop/controller.py`: `rotate_thumbnails` / `flip_thumbnails` turn the cached thumbnail bitmap in place (memory and disk, independently — they can be out of sync).
- `desktop/view/canvas/toolbar.py`: `rotate()`/`flip()` decide whether the active frame is actually part of the target set, then orchestrate the above.

## Test plan

- [x] `make lint` / `make type` clean
- [x] New unit tests for the transform, the active/other fan-out, and thumbnail turning
- [x] Full suite green (one pre-existing, unrelated macOS-only flake, also fails on `main`)
- [x] Manually verified in the app: batch rotate, batch flip, live thumbnail updates, deselecting the active frame before rotating leaves it untouched, and a batch rotate survives an app restart


thetalkingdrum:
Thanks for flagging this on the #1107 thread — reproduced and fixed in 51dcf09.

Root cause: `rotate_thumbnails`/`flip_thumbnails` only turn a thumbnail already cached in memory or on disk. A frame `generate_missing_thumbnails` is still decoding has neither yet, so the turn was silently dropped. `get_thumbnail_worker` never reads saved geometry either, and it persists its result to the disk cache before returning — so if that decode lands after the rotate, the old-orientation result gets written to disk and served back verbatim on every later request (the disk-cache check short-circuits the decode), sticking permanently until the frame is opened.

Fix: when a rotate/flip finds no disk-cached thumbnail yet, it queues the turn in a new `_thumbnail_pending_correction` map instead of dropping it. `_apply_thumbnails` replays any queued turn onto that decode's own delivery before committing it to memory and disk. Pruned against the live roll so a removed-then-re-added file sharing the same content hash can't inherit a stale queued turn.

Added tests for the race itself, that the correction only replays once, and that it's dropped if the frame leaves the roll first before the decode lands.

PR #1127, opened 2026-09-17
feat(thumbnails): background thumbnail refresh + manual Update Thumbnails control

thetalkingdrum:
## Summary

Applying settings, a preset, a scan-setup change, or Batch Analysis to more than the
current frame writes straight to the DB for every other touched frame — but nothing
refreshed those frames' filmstrip thumbnails short of opening them by hand.

- A background pass re-renders and persists each touched frame's thumbnail without
  opening it.
- It never blocks Export, Stitch, HDR, or anything else — no progress dialog pops, no
  batch is refused because of it.
- Auto Crop All and Batch Analysis do share its render thread, so they pre-empt it
  immediately (within about one frame) instead of waiting it out — the frames that got
  cut short are automatically retried once they're done, with a status message so it's
  not silent.
- Diptych/half-frame rows are intentionally left out of scope for now: a whole-frame
  render can't reproduce how their canvas joins two half configs.
- A manual **Update Thumbnails** toolbar button (whole roll) and context-menu entry
  (current selection) let you force this on demand, for staleness an automatic trigger
  missed or predates one. The same control turns into **Cancel** while it's running —
  a very large accidentally-opened folder needs a way out, and unlike a real batch's
  pre-emption, cancelling discards the backlog outright instead of resuming it.

## How it works

- `desktop/workers/render.py`: new `ThumbnailRenderWorker`, its own CPU-only
  `ImageProcessor` and decode cache, on the existing `norm_thread`.
- `desktop/session.py` / `desktop/controller.py`: a `frames_edited_offscreen` signal
  fired by every bulk-write path (Apply Settings, Apply Preset, scan setup, Batch
  Analysis, Apply Normalization Roll), dispatching `refresh_thumbnails_for`.
- `_begin_batch` pre-empts a running refresh only for the two batch types that actually
  share its thread (Auto Crop All, Batch Analysis) — everything else runs alongside it
  untouched.
- `desktop/view/sidebar/files.py`: the manual toolbar button and context-menu entry,
  routed through new `request_thumbnail_refresh`/`cancel_thumbnail_refresh` controller
  methods — same `refresh_thumbnails_for` plumbing underneath, no new backend path.

## Heads up on overlapping open PRs

- **#1118** ("Reset Roll to Defaults", now ready for review too) is a real, literal
  conflict: it inserts `reset_roll_settings` in `session.py` right after
  `apply_preset_fields`, the exact function this PR adds a
  `frames_edited_offscreen.emit(...)` block to the tail of — same hunk. It also edits the
  identical `docs/USER_GUIDE.md` right-click-menu sentence this PR does. Both are trivial
  to resolve by keeping both additions; whichever of us merges second does it by hand.
  Separately: this PR's `frames_edited_offscreen` signal is wired into Apply Settings,
  Apply Preset, scan setup, Batch Analysis, and Apply Normalization Roll, but not into
  #1118's `reset_roll_settings` — so a roll-wide or selection reset still leaves non-active
  frames' thumbnails stale after both land, unless `reset_roll_settings` is later wired to
  emit it too (or #1115's stale-dot approach lands and catches it structurally instead).

## Test plan

- [x] `make lint` / `make type` clean
- [x] New unit tests for the worker, the controller wiring, every bulk-write trigger's
      signal emit, the manual control's dispatch/cancel behavior, and the real widget
      wiring (toolbar button, context menu)
- [x] Full suite green (one pre-existing, unrelated Norwegian-keyboard tooltip failure,
      also fails on `main`)
- [x] Manually verified in the app: Apply Settings to roll, Apply Preset to roll,
      exporting a frame mid-refresh, Batch Analysis pre-empting and resuming, opening a
      different non-active frame mid-refresh, a forced-Slide-mode frame's color
      surviving the background render unchanged, the manual Update Thumbnails button
      and context-menu entry, and cancelling a roll-wide run mid-flight


thetalkingdrum:
Heads up: merge #1118 before this one. It touches the same hunk in `session.py` (`reset_roll_settings` lands right where this PR's `frames_edited_offscreen.emit(...)` goes, at the tail of `apply_preset_fields`) and the same `docs/USER_GUIDE.md` right-click-menu sentence.

When this rebases past #1118, fold `reset_roll_settings` into the wiring list here too. #1118 flagged it as a known limitation — a roll/selection reset leaves a stale thumbnail on non-active frames, the same gap Apply Settings/Apply Preset had before this PR — but the reset path itself isn't wired to emit the signal yet.

thetalkingdrum:
Thanks for the note on the #1107 thread about coordinating with #1121's memory work — added in 51af764.

GPU needed no coordination: `ThumbnailRenderWorker` already uses its own `ImageProcessor(use_gpu=False)`, so it never touches the GPU texture pool #1121 bounds.

RAM was the real gap: this refresh gets its own `PreviewManager` (following the same pattern the existing Auto Crop All worker already uses), so worst-case combined preview-cache footprint goes from 2x to 3x the fixed per-instance cap — and unlike #1121's prefetch path, nothing here checked memory pressure before starting a generation.

Fix: `refresh_thumbnails_for` now checks `available_system_memory_bytes()` against the same `MIN_RAM_RESERVE_BYTES` reserve #1121's prefetch admission uses (promoted out of `prefetch_policy`'s private constant so both share one number), and defers + retries after 5s instead of starting a refresh under pressure. Falls back to proceeding if the OS can't report available memory, so it doesn't silently stop working on platforms without that introspection.

